PR #17, opened 2021-06-02
SSV-Token-DEX-token-swap

vadiminc:


contracts/DEX.sol
```diff
@@ -0,0 +1,25 @@
...
+
+contract DEX {
+    event CDTToSSVConverted(uint256 amount);
+
+    IERC20 public cdtToken;
+    IERC20 public ssvToken;
+
+    uint public rate;
+
+    function init(address _cdtTokenAddress, address _ssvTokenAddress) public {
```
adamzigdon, line 13: you can change the parameters to be of type `IERC20` and then you won't need to cast in the following lines.

contracts/DEX.sol
```diff
@@ -0,0 +1,25 @@
+//SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.2;
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+contract DEX {
+    event CDTToSSVConverted(uint256 amount);
+
+    IERC20 public cdtToken;
+    IERC20 public ssvToken;
```
adamzigdon, line 9: if we want this contract to be generic you should rename the tokens variables to be `sourceToken` and `destinationToken`/`targetToken`.

contracts/DEX.sol
```diff
@@ -0,0 +1,25 @@
+//SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.2;
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+contract DEX {
+    event CDTToSSVConverted(uint256 amount);
```
adamzigdon, line 6: we should think if an event is really necessary here.
if so, i would remove the token names here and make it more general.
also, we should add the user address and have both amounts (source and target).

contracts/DEX.sol
```diff
@@ -0,0 +1,25 @@
...
+
+    uint public rate;
+
+    function init(address _cdtTokenAddress, address _ssvTokenAddress) public {
+        cdtToken = IERC20(_cdtTokenAddress);
+        ssvToken = IERC20(_ssvTokenAddress);
+        rate = 10;
+    }
+
+    function convertCDTToSSV(uint256 amount) public {
```
adamzigdon, line 19: rename this method as well.

contracts/SSVToken.sol
```diff
@@ -0,0 +1,9 @@
+//SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.2;
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+contract SSVToken is ERC20 {
```
adamzigdon, line 5: this file shouldn't be here.
if you want to have a token for testing it you should create a `mocks` folder.

contracts/tmp/OLDToken.sol
```diff
@@ -0,0 +1,9 @@
+//SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.2;
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+contract OLDToken is ERC20 {
```
adamzigdon, line 5: same here.

hardhat.config.ts
```diff
@@ -40,7 +40,13 @@ const config = {
       }
     ],
   },
-  networks: {},
+  networks: {
+    ganashe: {
```
adamzigdon, line 44: this is misspelled (ganache).

hardhat.config.ts
```diff
@@ -40,7 +40,13 @@ const config = {
...
+      url: 'http://127.0.0.1:7545',
+      accounts: [`0x15da52959efa15bc7a7b3c7bc293425e2e493a626a7c51e77b2b4326adcbe822`],
```
adamzigdon, line 46: that shouldn't be hardcoded (use env instead).

contracts/tmp/OLDToken.sol
adamzigdon: also why is `OLD` in capital letters?

hardhat.config.ts
```diff
@@ -52,7 +58,7 @@ if (process.env.GOERLI_ETH_NODE_URL) {
   config.networks['goerli'] = {
     url: process.env.GOERLI_ETH_NODE_URL,
     accounts: [`0x${process.env.GOERLI_OWNER_PRIVATE_KEY}`],
-    gasPrice: 98000000000
+    gasPrice: 320000000000
```
adamzigdon, line 61: i think that we should have a way to overwrite it via env and make the default smaller (320 gwei is very high).

scripts/dex-deploy.ts
```diff
@@ -0,0 +1,29 @@
...
+  const [deployer] = await ethers.getSigners();
+  console.log(
+    'Deploying the contracts with the account:',
+    await deployer.getAddress()
+  );
+
+  console.log('Account balance:', (await deployer.getBalance()).toString());
+
+  const Contract = await ethers.getContractFactory('DEX');
+  const contract = await upgrades.deployProxy(
```
adamzigdon, line 14: it's a very bad practice to have two variables that differ only in their case.

scripts/ssv-token-deploy.ts
```diff
@@ -0,0 +1,25 @@
+import { ethers } from 'hardhat';
+
+async function main() {
```
adamzigdon, line 3: mocks are used only for test purposes, they should be moved to test folder.

adamzigdon (changes requested):
see comments

contracts/DEX.sol
vadiminc: i think no, Lior wants to have two ways swap

contracts/DEX.sol
vadiminc: done

contracts/DEX.sol
```diff
@@ -0,0 +1,25 @@
...
+
+    IERC20 public cdtToken;
+    IERC20 public ssvToken;
+
+    uint public rate;
+
+    function init(address _cdtTokenAddress, address _ssvTokenAddress) public {
+        cdtToken = IERC20(_cdtTokenAddress);
+        ssvToken = IERC20(_ssvTokenAddress);
+        rate = 10;
```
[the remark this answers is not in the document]
vadiminc: done

contracts/DEX.sol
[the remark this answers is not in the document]
vadiminc: done

contracts/DEX.sol
vadiminc: will need to discuss with Lior currently just for 2 our tokens

contracts/SSVToken.sol
vadiminc: moved

contracts/tmp/OLDToken.sol
vadiminc: moved, just example

hardhat.config.ts
vadiminc: changed

hardhat.config.ts
vadiminc: fixed

hardhat.config.ts
vadiminc: ok

scripts/dex-deploy.ts
vadiminc: agreed, changed

scripts/ssv-token-deploy.ts
vadiminc: done

contracts/DEX.sol
lior-blox: Not sure we need an event

PR #18, opened 2021-06-09
dex improvements

vadiminc:


contracts/DEX.sol
```diff
@@ -9,6 +9,8 @@ contract DEX {
     event CDTToSSVConverted(uint256 amount);
     event SSVToCDTConverted(uint256 amount);
```
adamzigdon, line 10: I referred to these events in the previous PR and I didn't see a change.
It's very unclear what `amount` represents in each event.
The relevant wallet is not specified.
In addition, it can be inferred from other events within the transaction (Transfer).

contracts/DEX.sol
```diff
@@ -24,13 +26,19 @@ contract DEX {
         uint256 ssvAmount = amount / rate;
         cdtToken.transferFrom(msg.sender, address(this), amount);
```
adamzigdon, line 27: Instead of transferring `amount`, we should transfer `ssvAmount * rate`.
This way we're not locking remainders in the contract.
e.g.
rate = 10
user sends 15 CDT tokens, gets one SSV token in exchange but still sends 15.

contracts/DEX.sol
```diff
@@ -24,13 +26,19 @@ contract DEX {
...
         ssvToken.transfer(msg.sender, ssvAmount);
+        cdtAllowance[msg.sender] += amount;
```
adamzigdon, line 29: same here.
you can use a variable for it.

contracts/DEX.sol
```diff
@@ -24,13 +26,19 @@ contract DEX {
...
         emit CDTToSSVConverted(ssvAmount);
     }
 
     function convertSSVToCDT(uint256 amount) public {
         uint256 cdtAmount = amount * rate;
+        require(
+            cdtAllowance[msg.sender] >= cdtAmount,
+            "You can't get back amount of CDT tokens more than exchanged before"
```
adamzigdon, line 37: I would use a more general message (not using the word "You").

adamzigdon (changes requested):
In addition to the inline comments, you can remove the references to SafeMath.
You don't use it in the contract and it's redundant since Solidity 0.8.0.

contracts/DEX.sol
vadiminc: thank you, fixed

contracts/DEX.sol
vadiminc: fixed

contracts/DEX.sol
vadiminc: fixed

contracts/DEX.sol
vadiminc: fixed

PR #81, opened 2022-08-10
wip

vadiminc:


test/unit/ssv-registry.js
```diff
@@ -39,18 +39,75 @@ describe("Validators", () => {
...
-                sharePKs.slice(0, 4),
-                encryptedShares.slice(0, 4),
-                "10000"
-            ))
-            .to.emit(deployedRegistryContract, 'ValidatorAdded');
-            //.withArgs(validatorPK);
-
+        await progressBlocks(97);
+        const resultRegister = (await (await deployedRegistryContract.registerValidator(
+            [1, 2],
```
lior-blox, line 51: @vadiminc 
a. why only 2 operators
b. there are no check operators that exist?
c. please in tests have 2 registered validator cases:
1. (new collection and group creation) - most expensive
2. validator to the same/existing group - least expensive

PR #94, opened 2022-09-06
contract comments

andrew-blox:


contracts/SSVNetwork.sol
```diff
@@ -146,8 +146,8 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
                 ++dao.validatorCount;
                 _dao = dao;
             }
-
-            require(!_liquidatable(pod.balance, pod.validatorCount, operatorIds), "account liquidatable");
+            // IMPROPER ERROR SCHEMA ALSO WHY NOT IN  _validateValidatorParams FUNCTION
```
vadiminc, line 149: @andrew-blox there are different types of validation.
- validation for the input func params
- business logic validation (for instance, is liquidatable or not)

thats not good idea to merge them into one place.

contracts/SSVNetwork.sol
```diff
@@ -255,6 +255,7 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
         emit ValidatorRemoved(validatorPK, clusterId);
     }
 
+    // WHY NO VALIDATION ON AMOUNT?  WHAT IF NEW POD IS TOO EXPENSIVE AND DOES IMMEDIATE LIQUIDATION OR EVEN SKIPS LIQUIDATION AND GOES TO NEGATIVE
```
vadiminc, line 258: @andrew-blox  i think transfer is not completed logic, no sense to review it, it's just first impl version

contracts/SSVNetwork.sol
```diff
@@ -279,6 +280,8 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
             }
             // Changing to a single event reducing by 15K gas
         }
+
+        // WHY THIS EMIT IS NOT LAST THING OR AT LEAST AFTER VALIDATION
```
vadiminc, line 284: @andrew-blox in solidity it doesn't matter - if the validation will be failed, the func flow will be reverted

andrew-blox: Yeah @lior-blox updated me as well on that, i personally still think its best to keep last just to follow same format as other functions.

PR #112, opened 2022-11-13
core and tests changes for register validator

vadiminc:


contracts/SSVNetwork.sol
```diff
@@ -724,6 +727,24 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
...
+
+        bytes32 hashedPod = keccak256(abi.encodePacked(msg.sender, clusterId));
+
+        if (_pods[hashedPod].usage.block == 0) {
+            revert PodNotExists();
+        }
+
+        return clusterId;
+    }
+
```
andrew-blox, line 747: @lior-blox 
me personally, i wouldnt have the two reverts, mostly because its complicated for user
i would personally either remove the CluserNotExists validation or change the revert to PodNotExists

andrew-blox (approved):

andrew-blox:
General sidenote:
I suggested we start to use this schema to save gas bytes32 hashedPod = keccak256(abi.encodePacked(msg.sender, clusterId));

But it only saves gas if we call it multiple times, if you use var hashedPod only one time we actually spent more gas to create the memory of it for no reason

contracts/SSVNetwork.sol
```diff
@@ -724,6 +727,18 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
...
+    function getPod(uint64[] memory operatorIds) external view override returns(bytes32) {
+        _validateOperatorIds(operatorIds);
+
+        bytes32 clusterId = keccak256(abi.encodePacked(operatorIds));
+
+        if (_pods[keccak256(abi.encodePacked(msg.sender, clusterId))].usage.block == 0) {
+            revert PodNotExists();
+        }
+
+        return clusterId;
```
meshin-blox, line 739: How does this returns `clusterId` if method's name is `getPod`?
Shouldn't this method be called `getCluserId(operatorIds)`?
Because it really returns nothing related to pod

contracts/SSVNetwork.sol
lior-blox: Fixed by @andrew-blox suggestion

contracts/SSVNetwork.sol
lior-blox: We will review usage of pods vs cluster in the upcoming versions

lior-blox (approved):

PR #129, opened 2022-12-13
wip tests

vadiminc:


andrew-blox:
One major problem i see:
if operator is removed you cant register more validators to this pod (validation of operator amount wont pass), it kinda locks your pod, this also applies for remove

Basically if an operator removes himself and your pod uses him, you are locked.

andrew-blox:
There is a problem when you delete the operator
`delete _operators[operatorId];`
This will remove the operators snapshot also which will affect pod snapshot down the line

contracts/SSVNetwork.sol
```diff
@@ -239,413 +239,271 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
...
 
-        _deposit(msg.sender, clusterId, amount.shrink());
+        emit ValidatorRemoved(msg.sender, operatorIds, publicKey);
+        emit PodMetadataUpdated(msg.sender, operatorIds, pod);
     }
-    */
 
-    function liquidate(
-        address ownerAddress,
+    function liquidatePod(
```
andrew-blox, line 359: Maybe there should be a validation of pod.disabled with revert of podAlreadyLiquidated?

I get that it will revert with _liquidatable check, just a suggestion

andrew-blox (changes requested):

contracts/SSVNetwork.sol
```diff
@@ -239,413 +239,271 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
...
-    /*
-    function deposit(address owner, bytes32 clusterId, uint256 amount) external override {
-        _validateClusterId(clusterId);
+    function deposit(
+        address owner,
+        uint64[] memory operatorIds,
+        uint256 amount,
+        Pod memory pod
+    ) external override {
+        _validatePodIsNotLiquidated(pod);
```
andrew-blox, line 473: Not sure this is correct.
I think you should be allowed to deposit to a liquidated pod.

contracts/SSVNetwork.sol
```diff
@@ -687,37 +545,49 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
...
 
-    /*
-    function withdrawPodBalance(bytes32 clusterId, uint256 amount) external override {
-        _validateClusterId(clusterId);
+    function withdrawPodBalance(
+        uint64[] memory operatorIds,
+        uint256 amount,
+        Pod memory pod
+    ) external override {
+        _validatePodIsNotLiquidated(pod);
```
andrew-blox, line 553: If we start to allow deposit to liquidated pod then we must remove this validation

contracts/SSVNetwork.sol
vadiminc: @andrew-blox no that's not. we have validation in function _validatePodIsNotLiquidated(pod);

contracts/SSVNetwork.sol
vadiminc: no, if you wanna deposit there is a function reactivatePod with deposit and important logic for that like podIndex recalculation

PR #134, opened 2022-12-20
fee recipient logic

vadiminc:


contracts/SSVNetwork.sol
```diff
@@ -654,6 +654,10 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
         return (feeChangeRequest.fee.expand(), feeChangeRequest.approvalBeginTime, feeChangeRequest.approvalEndTime);
     }
 
+    function feeRecipientAddress(address recipientAddress) external override {
```
andrew-blox, line 657: in iSSV file you define this param as feeRecipientAddress but here as recipientAddress
Either name is fine but how does this actually work with mismatched names?

andrew-blox (changes requested):

andrew-blox:
Maybe add a test so we keep full coverage

contracts/SSVNetwork.sol
vadiminc: it doesn't matter.

vadiminc:
> Maybe add a test so we keep full coverage

it has the test already

contracts/SSVNetwork.sol
vadiminc: oh, the file is new, forgot to add into the pr. check now

mtabasco, line 657: not sure if this function should be called only from EOA or can be called from contracts too.
if its mandatory only for the former, then you can check:
```
if(msg.sender == tx.origin) {
    revert OnlyEOAAllowed();
}
```
cc @andrew-blox

andrew-blox: I dont think its needed here but thanks @mtabasco for this article it helped explain the concept to me:
https://ethereum.stackexchange.com/questions/113962/what-does-msg-sender-tx-origin-actually-do-why

andrew-blox (approved):

mtabasco (approved):

contracts/ISSVNetwork.sol
```diff
@@ -131,6 +131,11 @@ interface ISSVNetwork {
         Pod pod
     );
 
+    event FeeRecipientAddressAdded(
+        address ownerAddress,
```
lior-blox, line 135: @vadiminc @mtabasco Should this be indexed address?

mtabasco: It depends on the use case. If we want to allow search using `ownerAddress`, `recipientAddress` or both, yes, then should be indexed

PR #138, opened 2022-12-21
Update shares and combine emits io1 2131

andrew-blox:


contracts/SSVNetwork.sol
```diff
@@ -302,8 +303,8 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
 
         _pods[hashedPod] = keccak256(abi.encodePacked(pod.validatorCount, pod.networkFee, pod.networkFeeIndex, pod.index, pod.balance, pod.disabled ));
 
-        emit ValidatorAdded(msg.sender, operatorIds, publicKey, shares);
-        emit PodMetadataUpdated(msg.sender, operatorIds, pod);
+        emit ValidatorAdded(msg.sender, operatorIds, publicKey, sharePublicKeys, encryptedKeys, pod);
```
mtabasco, line 306: the length of `operatorIds` should match with `sharePublicKeys` and `encryptedKeys`, isn't it? 
depending on who reads this event, a check `operatorIds.length == sharePublicKeys.length` may be but in place

PR #151, opened 2022-12-26
operator removal logic

vadiminc:


andrew-blox (changes requested):

andrew-blox:
I just wanna add that with this logic if user has 5 validators in a pod with operators 1,2,3,4, then operator 4 removes himself.  User is now locked to add more validators to this pod (he will revert with operator doesnt exist).

I am talking with @arielblox to see if this is ok.  If not maybe a solution is to check if its first register or not
if first time registering revert if operator not exist if not first time then dont revert but lets see what Ariel says.

andrew-blox (approved):

PR #160, opened 2023-01-02
list of gas optz changes IO1-1433

vadiminc:
Link to document:
[https://docs.google.com/document/d/1UONeUat6DFKZI-E9aiG0eKkhShamU2q3g0KAsmC6BKI](https://docs.google.com/document/d/1UONeUat6DFKZI-E9aiG0eKkhShamU2q3g0KAsmC6BKI)

- [x] for loops
- [x] Return calculated values from private functions to not recalculate / access them again and to avoid repetitive checks
- [x] Nested if
- [x] Use calldata in function parameters when possible (see comments in doc, although it can represent some savings in specific functions, overall it doesn't represent a gain)
- [-] No need to initialize variables with default values
- [x] Follow check-effects-interaction pattern
- [x] No repeated computations
- [x] Check proper copy of storage structs to memory

| function  | contract-v3  | v3-gas-optimizations | savings |
|---|---|---|---|
| liquidatePod  | 131499  |  129601 | 1898  |
| reactivatePod  |  125151 |  124879 | 272 |  
| registerValidator  |  214782 |  213831 | 951 |   
| removeValidator  | 96911 |  96440 | 471 | 
| withdrawPodBalance |  91924|  89996 | 1928 |

### DIFF
--- a/contracts/SSVNetwork.sol
+++ b/contracts/SSVNetwork.sol
@@ -235,7 +235,7 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
     /********************************/
     function registerValidator(
         bytes calldata publicKey,
-        uint64[] calldata operatorIds,
+        uint64[] memory operatorIds,
         bytes calldata shares,
         uint256 amount,
         Pod memory pod
@@ -315,7 +315,7 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
 
     function removeValidator(
         bytes calldata publicKey,
-        uint64[] calldata operatorIds,
+        uint64[] memory operatorIds,
         Pod memory pod
     ) external override {
         uint operatorsLength = operatorIds.length;
@@ -368,7 +368,7 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
 
     function liquidatePod(
         address owner,
-        uint64[] calldata operatorIds,
+        uint64[] memory operatorIds,
         Pod memory pod
     ) external override {
         _validatePodIsNotLiquidated(pod);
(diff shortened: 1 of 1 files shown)


contracts/SSVNetwork.sol
```diff
@@ -239,13 +235,15 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
     /********************************/
     function registerValidator(
         bytes calldata publicKey,
-        uint64[] memory operatorIds,
+        uint64[] calldata operatorIds,
```
mtabasco, line 238: @vadiminc I remember applying calldata _everywhere_ doesnt reduce gas, only with certain functions. Have you tested it represent savings where its applied? thanks

mtabasco:
@vadiminc can you please update the table in the initial comment to see the progress of the numbers with
`contract-v3` vs `v3-gas-optimizations` in terms of gas? thanks

contracts/SSVNetwork.sol
```diff
@@ -286,11 +288,7 @@ contract SSVNetwork is OwnableUpgradeable, ISSVNetwork {
             }
         }
 
-        if (amount > 0) {
-            _deposit(msg.sender, operatorIds, amount.shrink());
-            pod.balance += amount.shrink();
-        }
-
+        pod.balance += amount.shrink();
```
andrew-blox, line 291: Im pretty sure that if you send an amount of 0 this will cause the tx to fail right? shrink() calls shrinkable()

  function shrinkable(uint256 value) internal pure returns (uint256) {
    require(value % DEDUCTED_DIGITS == 0, "Precision is over the maximum defined");
    return value;
  }

andrew-blox (changes requested):

andrew-blox (approved):

mtabasco (approved):

PR #162, opened 2023-01-05
extra func logic

vadiminc:


andrew-blox:
Main thing is regarding ticket https://bloxxx.atlassian.net/browse/IO1-2199: Remove internal deposit emit

I see you did remove which is good but please update all the places with relevent amount emits.
For example in registerValidator you have validatorAdded event, it should emit the amount as well.  An ex of other place is reactivatePod emit should have the amount etc..  All places that uses internal deposit should emit some sort of amount.

PR #177, opened 2023-01-30
Add slither to CI pipeline - IO1-2233 

mtabasco:


.github/workflows/slither.yml
```diff
@@ -0,0 +1,16 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Run Slither
+        uses: crytic/slither-action@v0.2.0
+        id: slither
+        with:
+          node-version: 16
+          fail-on: high
```
mtabasco, line 15: @andrew-blox Fail when high severity issues found. Although all the issues should be checked (medium and low), there could be some false positives that prevent the task from being completed. Tweaking the code to skip them will add unnecessary noise.

.github/workflows/slither.yml
```diff
@@ -0,0 +1,16 @@
...
+          slither-args: --exclude controlled-delegatecall
```
mtabasco, line 16: @andrew-blox excluding this issue as it belongs to OZ's proxy implementation

.github/workflows/slither.yml
andrew-blox: as we talked maybe down the road a better solution will be to put on low and exclude whats needed in the slither-args instead

andrew-blox (approved):

lior-blox (approved):

PR #178, opened 2023-01-31
Limit max # validators per operator - IO1-2160

mtabasco:
Provide functionality to modify the value via upgrades


contracts/test/SSVNetwork_v2.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity 0.8.16;
+
+import "../SSVNetwork.sol";
+
+contract SSVNetwork_v2 is SSVNetwork {
+    
+    function initializev2(uint32 validatorsPerOperatorLimit_) reinitializer(2) external {
```
andrew-blox, line 8: i suggest making a var called version and instead of hard code 2 to show version + 1 so user better understands how to handle this but either way is fine

test/helpers/contract-helpers.ts
```diff
@@ -156,19 +156,58 @@ export const registerValidators = async (ownerId: number, numberOfValidators: nu
...
     args = result.eventsByName.ValidatorAdded[0].args;
-
     DB.validators.push({ publicKey, operatorIds, shares });
     validators.push({ publicKey, shares });
   }
 
   return { validators, args };
 };
 
+export const registerValidatorsRaw = async (ownerId: number, numberOfValidators: number, amount: string, operatorIds: number[]) => {
```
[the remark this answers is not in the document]
andrew-blox: awesome!!  Later we should update in relevant places, it will defintely cut down lines of code in the stress test ;)

andrew-blox (approved):

contracts/test/SSVNetwork_v2.sol
mtabasco: cool, let me merge first the _version function_ PR and adapt this test to make it coherent as you suggest, thanks!

lior-blox (approved):

PR #179, opened 2023-02-01
Versioned contract - IO1-1818

mtabasco:
Create a public `version` state variable that can be queried by anyone anytime but only upgraded via the upgrade process.


andrew-blox (approved):

lior-blox:
@mtabasco lets use a string version such as 1.0.0 (this would be the first for mainnet)

lior-blox (approved):

PR #185, opened 2023-02-12
Network fee math - IO1-2028

mtabasco:
DAO struct simplified:

```
struct DAO {
    uint32 validatorCount;
    uint64 balance;
    uint64 block; // keeps track of the latest block when the balance is updated
}
```


mtabasco:
**Required test**
Setup:
6 operators
1 validator 1,2,3,4
2 validator 3,4,5,6

**Execution**

1. progressBlocks
2. Increase network fee
3. progressBlocks
4. Decrease network fee

Check balances during the process

mtabasco:
> **Required test** Setup: 6 operators 1st validator 1,2,3,4 2nd validator 3,4,5,6
> 
> **Execution**
> 
> 1. progressBlocks
> 2. Increase network fee
> 3. progressBlocks
> 4. Decrease network fee
> 
> Check balances during the process

 @andrew-blox Done

andrew-blox (approved):
The math is a little above me to be honest.  The new balance test is looking a bit better to make sure that we catch such an event but i think it would be best if product also takes a look at this PR.  

I dont quite fully understand if network fee has an index, looks to be no which i then question are we always with the correct balances even after a network fee change but the tests seem to support it.

I will approve but i would feel better with it if product reviewed this as well.

mtabasco:
> The math is a little above me to be honest. The new balance test is looking a bit better to make sure that we catch such an event but i think it would be best if product also takes a look at this PR.
> 
> I dont quite fully understand if network fee has an index, looks to be no which i then question are we always with the correct balances even after a network fee change but the tests seem to support it.
> 
> I will approve but i would feel better with it if product reviewed this as well.

@andrew-blox Thanks! sure, I will ask Product and will explain the changes.

lior-blox (approved):

PR #187, opened 2023-02-15
Emit non shrunk amounts in events - IO1-2357

mtabasco:
Check all amount inputs in functions and outputs in events are `uint256` types.
This doesn't apply for `declareOperatorFeePeriod`, `executeOperatorFeePeriod` and `operatorMaxFeeIncrease` that still keep `uint64` type


contracts/SSVNetwork.sol
```diff
@@ -318,7 +318,7 @@ contract SSVNetwork is UUPSUpgradeable, OwnableUpgradeable, ISSVNetwork {
             network_
         );
 
-        cluster.balance += amount.shrink();
+        cluster.balance += amount;
```
andrew-blox, line 321: Something seems off here.
cluster.balance is a shrunk amount right?
shouldnt it be cluster.balance.expand() += amount?
ahhh ok i see you never save it as shrunk so its always expanded?  WHat about if i do direct deposit or direct withdraw are we using expanded amounts?

Ok i checked the file as a whole seems ok i see you updated deposit and withdraw as well.

@lior-blox do you think its worth @mtabasco to see what gas costs will be if we removed shrink and expand from operators fee as well?  As of now this is the only reason we use these shrink / expand functions / precision.

andrew-blox (approved):

lior-blox:
duplicate pr

PR #192, opened 2023-02-25
Fix event typo

rotcivegaf:


lior-blox:
@rotcivegaf Closing as its we work on v3 contracts which refactors this function

PR #198, opened 2023-03-03
Liquidate cluster negative balance - IO1-2756

mtabasco:
- A cluster with negative balance can be liquidated instead of returning `InssuficientFunds` error.
The remaining balance is transferred to the liquidator and the owed funds are lost.

- Checks on other functions: `registerValidator`, `removeValidator`, `reactivate`.
- Refactor ClusterLib.
- Udgrade solidity and hardhat version.
- Linter


contracts/SSVNetwork.sol
```diff
@@ -194,41 +180,35 @@ contract SSVNetwork is UUPSUpgradeable, Ownable2StepUpgradeable, ISSVNetwork {
...
 
-        bytes32 hashedCluster = keccak256(
-            abi.encodePacked(msg.sender, operatorIds)
-        );
-        {
-            bytes32 hashedClusterData = keccak256(
+        bytes32 hashedCluster = keccak256(abi.encodePacked(msg.sender, operatorIds));
+
+        if (clusters[hashedCluster] == bytes32(0)) {
+            cluster = Cluster({validatorCount: 0, networkFeeIndex: 0, index: 0, balance: 0, active: true});
```
andrew-blox, line 189: @lior-blox Do we want to force active: true here?  If the user registers first cluster with false we will revert it to hardcoded true

andrew-blox (changes requested):

contracts/SSVNetwork.sol
andrew-blox: Actually maybe this is better solution

if (clusters[hashedCluster] == bytes32(0)) {
            if (!cluster.active) {
                revert ActiveMustBeTrue
            }
            cluster = Cluster({validatorCount: 0, networkFeeIndex: 0, index: 0, balance: 0, active: true});
        }

andrew-blox (approved):

PR #200, opened 2023-03-14
No timelocked reduce operator fee functionality - IO1-2834

mtabasco:
Create `reduceOperatorFee`. Execution restricted to the operator owner.


test/sanity/balances.ts
```diff
@@ -27,6 +27,8 @@ describe('Balance Tests', () => {
     // Set network fee
     await ssvNetworkContract.updateNetworkFee(networkFee);
 
+    console.log("1..", await ethers.provider.getBlockNumber());
```
drewwilliams11111, line 30: do we need these console.logs?

drewwilliams11111 (approved):

PR #201, opened 2023-03-14
SSVNetworkViews.getValidator function

mtabasco:


contracts/SSVNetwork.sol
```diff
@@ -45,7 +45,7 @@ contract SSVNetwork is UUPSUpgradeable, Ownable2StepUpgradeable, ISSVNetwork {
     mapping(uint64 => Operator) public operators;
     mapping(uint64 => OperatorFeeChangeRequest) public operatorFeeChangeRequests;
     mapping(bytes32 => bytes32) public clusters;
-    mapping(bytes32 => Validator) private _validatorPKs;
+    mapping(bytes32 => Validator) public _validatorPKs;
```
mtabasco, line 48: @andrew-blox storage layout compatible (upgrade safe), but a big TODO for changing the name without `_` before mainnet (or another testnet deployment).

andrew-blox: @mtabasco You think its better to do like this then to make a get function for it?

mtabasco: Mostly for 2 reasons:
1. don't increase size to SSVNetwork contract.
2. follow the pattern of putting all `get` functions in Views contract

lior-blox (approved):

PR #202, opened 2023-03-15
Implement Minimum Liquidation Collateral - IO1-2838

mtabasco:
New state variable: `minimumLiquidationCollateral` with `update` function associated.


contracts/SSVNetwork.sol
```diff
@@ -82,7 +83,8 @@ contract SSVNetwork is UUPSUpgradeable, Ownable2StepUpgradeable, ISSVNetwork {
         uint64 operatorMaxFeeIncrease_,
         uint64 declareOperatorFeePeriod_,
         uint64 executeOperatorFeePeriod_,
-        uint64 minimumBlocksBeforeLiquidation_
+        uint64 minimumBlocksBeforeLiquidation_,
+        uint64 minimumLiquidationCollateral_
```
mtabasco, line 87: If we want to take minimumLiquidationCollateral as input in the initializer function, its mandatory to use a separate contract with a new initializer. So then we can have the changes in current stage deployed contracts.
But for the audit and mainnet deployment, we can go with this implementation.

test/liquidate/reactivate.ts
```diff
@@ -61,7 +61,7 @@ describe('Reactivate Tests', () => {
     await expect(ssvNetworkContract.connect(helpers.DB.owners[1]).reactivate(updatedCluster.operatorIds, minDepositAmount, updatedCluster.cluster)).to.emit(ssvNetworkContract, 'ClusterReactivated');
   });
 
-  it('Reactivate with 0 deposit and no validators emits ClusterReactivated', async () => {
+  it.skip('Reactivate with 0 deposit and no validators emits ClusterReactivated', async () => {
```
mtabasco, line 64: @andrew-blox I think reactivating with 0 balance no longer makes sense, as the check for `isLiquidatable` will return true. Can a cluster with zero balance be reactivated?

andrew-blox: not sure, i mean if there are no validators in the cluster then i guess theoretically you can reactivate with 0 but no other way, even if you uahve 4 operators with 0 fee there will still be network foo

contracts/SSVNetwork.sol
```diff
@@ -612,6 +634,11 @@ contract SSVNetwork is UUPSUpgradeable, Ownable2StepUpgradeable, ISSVNetwork {
         emit LiquidationThresholdPeriodUpdated(blocks);
     }
 
+    function updateMinimumLiquidationCollateral(uint64 amount) external override onlyOwner {
+        minimumLiquidationCollateral = amount;
```
andrew-blox, line 638: shouldnt it be amount.shrink() ?

contracts/libraries/ClusterLib.sol
```diff
@@ -22,8 +22,11 @@ library ClusterLib {
         ISSVNetworkCore.Cluster memory cluster,
         uint64 burnRate,
         uint64 networkFee,
-        uint64 minimumBlocksBeforeLiquidation
+        uint64 minimumBlocksBeforeLiquidation,
+        uint64 minimumLiquidationCollateral
     ) internal pure returns (bool) {
+        if (cluster.balance < minimumLiquidationCollateral.expand()) return true;
```
andrew-blox, line 28: this is mainly why i point out the shrink
`minimumLiquidationCollateral.expand()`

andrew-blox (changes requested):

contracts/SSVNetwork.sol
```diff
@@ -611,6 +634,15 @@ contract SSVNetwork is UUPSUpgradeable, Ownable2StepUpgradeable, ISSVNetwork {
         emit LiquidationThresholdPeriodUpdated(blocks);
     }
 
+    function updateMinimumLiquidationCollateral(uint64 amount) external override onlyOwner {
+        if (amount < MINIMAL_LIQUIDATION_COLLATERAL) {
+            revert NewCollateralIsBelowMinimum();
+        }
```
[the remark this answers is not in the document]
andrew-blox: this is resolved

contracts/SSVNetwork.sol
```diff
@@ -103,14 +106,16 @@ contract SSVNetwork is UUPSUpgradeable, Ownable2StepUpgradeable, ISSVNetwork {
...
+        uint64 minimumBlocksBeforeLiquidation_,
+        uint64 minimumLiquidationCollateral_
     ) internal onlyInitializing {
         version = bytes32(abi.encodePacked(initialVersion_));
         _token = token_;
         operatorMaxFeeIncrease = operatorMaxFeeIncrease_;
         declareOperatorFeePeriod = declareOperatorFeePeriod_;
         executeOperatorFeePeriod = executeOperatorFeePeriod_;
         minimumBlocksBeforeLiquidation = minimumBlocksBeforeLiquidation_;
+        minimumLiquidationCollateral = minimumLiquidationCollateral_;
```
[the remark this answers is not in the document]
mtabasco: `minimumLiquidationCollateral_` is expected to be a _shrinked_ amount already, unless we want to pass uint256 (expanded) value.

contracts/libraries/ClusterLib.sol
mtabasco: yes, so `minimumLiquidationCollateral` should be expanded to be comparable to `cluster.balance` that is a _expanded_ value (uint256)

PR #203, opened 2023-03-15
Refactor update operators loop

mtabasco:
`reactivate`, `removeValidator`, `liquidate` functions have similiar loop over operators involved in the cluster.
Refactor it.


contracts/SSVNetwork.sol
```diff
@@ -807,6 +756,33 @@ contract SSVNetwork is UUPSUpgradeable, Ownable2StepUpgradeable, ISSVNetwork {
...
+                burnRate += operator.fee;
+                operators[operatorIds[i]] = operator;
+            }
+
+            clusterIndex += operator.snapshot.index;
+            unchecked {
+                ++i;
+            }
+        }
+    }
```
andrew-blox, line 784: shouldnt we have an additional line after the for loop with something like:

burnRate += networkFee * validatorCount

andrew-blox (changes requested):

contracts/SSVNetwork.sol
mtabasco: Good observation @andrew-blox , thanks. after calling `_updateOperators` , there is a call to either `ClusterLib.updateBalance` or `ClusterLib.updateClusterData` where the `networkFee` and `validatorCount` is taking into account to update the cluster's balance.

andrew-blox (approved):

PR #205, opened 2023-03-20
Publish NPM package with contracts

tuleyko:
As StakeStar we are building Liquid Staking pool based on SSV Network and obviously we are using your contracts under the hood. 

Your contracts aren't published anywhere via NPM packages, so we have to manually copy-paste them into our repo.

You are going to publish fresh contracts(v3) soon and we will have to copy-paste them again.

This pull-request adds required configuration to package.json to allow publishing to npm.pkg.github.com and easier integration for protocols. After merging, please consider setting correct version and publishing a package for every release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,10 @@
 {
-  "name": "ssv-network",
+  "name": "@bloxapp/ssv-network",
+  "version": "0.2.0",
+  "files": [
+    "/contracts/**/*.sol",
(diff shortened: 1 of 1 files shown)


mtabasco:
@tuleyko I'm closing this one because it is addressed here: https://github.com/bloxapp/ssv-network/pull/277
thanks!

PR #208, opened 2023-03-28
Audit tasks

mtabasco:
QS-3 Register Same Operators
QS-8 Same Operator Registered Multiple Times
~~QS-9 Missing Input Validation~~
QS-10 Type256.shrink() Can Overflow
QS-16 Incompatible with Deflationary Tokens (check)


contracts/SSVNetwork.sol
```diff
@@ -171,6 +170,7 @@ contract SSVNetwork is UUPSUpgradeable, Ownable2StepUpgradeable, ISSVNetwork {
     }
 
     function setFeeRecipientAddress(address recipientAddress) external override {
+        if (recipientAddress == address(0)) revert ZeroAddressNotAllowed();
```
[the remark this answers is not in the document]
andrew-blox: @lior-blox wht do you think.

contracts/SSVNetwork.sol
```diff
@@ -658,15 +665,23 @@ contract SSVNetwork is UUPSUpgradeable, Ownable2StepUpgradeable, ISSVNetwork {
...
-    }
 
-    function _validateOperatorIds(uint operatorsLength) private pure {
-        if (operatorsLength < 4 || operatorsLength > 13 || operatorsLength % 3 != 1) {
-            revert InvalidOperatorIdsLength();
+        validatorPk = keccak256(publicKey);
+        if (
+            validatorPk ==
+            keccak256(
+                hex"000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
```
andrew-blox, line 677: Also this one

andrew-blox (approved):

contracts/SSVNetwork.sol
```diff
@@ -143,6 +147,8 @@ contract SSVNetwork is UUPSUpgradeable, Ownable2StepUpgradeable, ISSVNetwork {
             validatorCount: 0,
             fee: fee.shrink()
         });
+        operatorsPKs[hashedPk] = 0x01;
```
andrew-blox, line 150: I personally think we should save the operator id here.  
Maybe we wont use it but we could have this data saved for the same cost as just "0x01" and maybe down the road we might want to know this data and it will already be there at the same cost.

andrew-blox (approved):

PR #209, opened 2023-03-29
Audit best practices

mtabasco:
- [x] Invoke _disableInitializers() on the constructor
- [x] visibility of the following state vairables should be set: 
        - SSVNetworkViews._ssvNetwork 
        - SSVNetworkViews.__gap 
        - SSVNetwork.__gap 
- [x] Remove updateClusterData in reactivate() 
- [x] removeValidator() validates the operator's length and validator public key with calls to _validateOperatorIds() and _validatePublicKey() . These checks are not needed
- [ ] consider using unchecked blocks around arithmetic that you are certain will not overflow (for example, ++cluster.validatorCount or ++dao_.validatorCount ).
- [x] ClusterLib.validateHashedCluster() calls ssvNetwork.clusters() twice to obtain the same hashed cluster data. Store the result of ssvNetwork.clusters() in a memory variable 
- [x] registerValidator() calls operatorsWhitelist twice to obtain the same whitelist address for a given operator. Consider storing the returned value in a memory variable 
- [x]  _withdrawOperatorEarnings() calls amount.shrink() twice to obtain the same value. Consider storing the returned value in a memory variable 
- [x] withdraw() changes operator from storage to memory. Check
- [x] In order to save storage slots in struct Cluster , consider ordering the fields such that the uint256 balance occurs after all the other fields.
- [x] error NoFeeDelcared() is misspelled and should be corrected NoFeeDeclared() .
- [x] uint64 currrentNetworkFeeIndex should be corrected uint64 currentNetworkFeeIndex
- [x] function _validateOperatorIds() should be renamed function _validateOperatorsLength() (removed)
- [ ] consider renaming the owner variable to avoid shadow conflicts
- [x] Consider replacing the usage of magic numbers with constant variables. 
    - 1. SSVNetwork.__SSVNetwork_init_unchained() : 2_000 
    - 2. SSVNetwork._declareOperatorFee() : 10000
- [x] ISSVNetwork.registerValidator.sharesEncrypted does not match SSVNetwork.registreValidator.shares


contracts/SSVNetwork.sol
```diff
@@ -458,8 +464,6 @@ contract SSVNetwork is UUPSUpgradeable, Ownable2StepUpgradeable, ISSVNetwork {
         cluster.index = clusterIndex;
         cluster.networkFeeIndex = currentNetworkFeeIndex;
 
-        cluster.updateClusterData(clusterIndex, currentNetworkFeeIndex);
```
mtabasco, line 461: Recheck this

andrew-blox (approved):

PR #211, opened 2023-04-03
Use SSVNetwork interface in SSVNetworkViews

mtabasco:
Use `ISSVNetwork` in SSVNetworkViews and libraries to facilitate integrations.


andrew-blox (approved):
Why did you implement like this?  Does it save gas, bring down contract size, or just a standard?

mtabasco:
> Why did you implement like this? Does it save gas, bring down contract size, or just a standard?

@andrew-blox initially this suggestion came from stakestar team and makes totally sense. 3 key points:
UX: interfaces provide a clear abstraction of the functionality without exposing any implementation details. This makes it easier for users to understand what the contract does. 
Security:  passing the interface as a parameter to the initializer / constructor is generally considered safer, mostly because interfaces only provide a limited set of functions that can be called, which reduces the attack surface of the contract.
Upgradeability: Not the main reason, but interfaces can be implemented by different contracts, which allows for more flexibility when upgrading the contract.

PR #212, opened 2023-04-13
Bulk liquidation - IO1-2852

mtabasco:
Implement bulk cluster liquidation.

- new `Liquidation` struct:
```
struct Liquidation {
        address owner;
        uint64[] operatorIds;
        Cluster cluster;
    }
```
- ~~`liquidate` function now accepts a list of `Liquidation` objects. This function is used for individual or multiple liquidations~~
- `liquidate` function now only liquidate one cluster. `bulkLiquidate` accepts a list of `Liquidation` objects.
- Premise: `operatorIds` in `Liquidation` must be ordered. This is checked implicitly when validating cluster data, as the operators must be ordered when registering a validator
- `bulkLiquidate` is optimized to read / write operators from / storage once per operator. The DAO and execution of the liquidation amount transfer is done once too.
- new event `ClustersLiquidated` that emits a list of `Liquidation` with updated clusters data.

Gas numbers:
| operators | clusters | validators | gas |
|---|---|---|---|
| 4 | 1 | 10 | 126100 |
| 6 | 2 | 18 | 200400 |

Difference between previous `liquidate` function and new one liquidating one cluster, same data:

| operators | clusters | validators | gas previous liquidate | gas new liquidate |
|---|---|---|---|---|
| 4 | 1 | 1 | 124895 | 126100 |


contracts/SSVNetwork.sol
```diff
@@ -359,71 +359,95 @@ contract SSVNetwork is UUPSUpgradeable, Ownable2StepUpgradeable, ISSVNetwork {
...
-                    burnRate += operator.fee;
-                    operators[operatorIds[i]] = operator;
-                }
+        uint64 networkFeeIndex = NetworkLib.currentNetworkFeeIndex(network);
 
-                clusterIndex += operator.snapshot.index;
-                unchecked {
-                    ++i;
-                }
+        for (uint i; i < liquidations.length; ++i) {
```
andrew-blox, line 372: I think we must have some number limit of validators to liquidate due to this for loop.  @lior-blox what do you think is a good number here?

andrew-blox:
I am not complete with the CR.  I stopped because I am not sure its the best implementation of it.  Maybe i can make a suggestion for it and if you dont agree with it let me know and I will do CR of current code?

I suggest a private liquidate function and 2 public functions of liquidate, maybe something like `liquidate()` and `bulkLiquidate()` and basically both will only call the private liquidate function but the bulk will have a for loop and array param with a limit of validators?

@lior-blox @mtabasco

lior-blox:
Closed as requires further thinking

PR #213, opened 2023-04-18
Bulk registration - IO1-2031

mtabasco:
- Register multiple validators using the same cluster
- Emit one event with all public keys and shares


lior-blox:
Closed as requires further thinking

PR #220, opened 2023-05-18
Withdraw fix

mtabasco:
The _liqudatable_ state of a cluster should be independent from deposit / withdraw actions.
In this PR, the check `isLiquidatable` is removed from `withdraw` function.


contracts/SSVNetwork.sol
```diff
@@ -524,17 +524,6 @@ contract SSVNetwork is UUPSUpgradeable, Ownable2StepUpgradeable, ISSVNetwork {
...
-            cluster.isLiquidatable(
-                burnRate,
-                network.networkFee,
-                minimumBlocksBeforeLiquidation,
-                minimumLiquidationCollateral
-            )
-        ) {
-            revert InsufficientBalance();
-        }
-
```
andrew-blox, line 537: This is very important to keep,
If you remove this then people can withdraw past there collateral zone

The one to remove is cluster.validateClusterIsNotLiquidated();
We dont care if you are liquidated already but we do care if you put yourself to a liquidatable place

andrew-blox (changes requested):

contracts/SSVNetwork.sol
lior-blox: and add a test for it

PR #228, opened 2023-05-30
Reset declared operator fee when reducing it

mtabasco:
If a new valid fee is declared for an operator and the fee is reduced, the declared operator's fee is cancelled.


andrew-blox (approved):

mtabasco:
Changes in this PR already implemented in `revamped` branch

PR #239, opened 2023-06-23
Fix client gas estimations

mtabasco:
Converge delegate functions into one. Don't use it from a library but from an abstract contract.
Update gas metrics.


andrew-blox (approved):

contracts/SSVProxy.sol
```diff
@@ -0,0 +1,32 @@
...
+
+            // Call the implementation.
+            // out and outsize are 0 because we don't know the size yet.
+            let result := delegatecall(gas(), implementation, 0, calldatasize(), 0, 0)
+
+            // Copy the returned data.
+            returndatacopy(0, 0, returndatasize())
+
+            switch result
+            // delegatecall returns 0 on error.
```
lior-blox, line 23: if this knows to return an error, then why ssv.network doesn't?

lior-blox (approved):

PR #249, opened 2023-07-13
DAO audit recommendations

mtabasco:
- [x] When using the Diamond Pattern, the gas consumption can be minimized by load storage pointers only once per function execution. The following functions access the same storage pointer more than once during a single function execution: SSVDAO.updateNetworkFee()

- [x] ProtocolLib.sol#L42 � The check > type(uint32).max is unnecessary -> Keep it because of the explicit revert we use VS default solidity one
- [x] SSV11 Inconsistent Use of shrink() and expand() 

**Upgrade plan**

Update SSVClusters module

### DIFF
--- a/contracts/modules/SSVDAO.sol
+++ b/contracts/modules/SSVDAO.sol
@@ -3,24 +3,21 @@ pragma solidity 0.8.18;
 
 import "../interfaces/ISSVDAO.sol";
 import "../libraries/Types.sol";
-import "../libraries/OperatorLib.sol";
 import "../libraries/ProtocolLib.sol";
-import "../libraries/CoreLib.sol";
-import "../libraries/SSVStorage.sol";
 
 contract SSVDAO is ISSVDAO {
     using Types64 for uint64;
     using Types256 for uint256;
 
     using ProtocolLib for StorageProtocol;
-    
+
     uint64 private constant MINIMAL_LIQUIDATION_THRESHOLD = 100_800;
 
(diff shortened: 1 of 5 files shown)


contracts/modules/SSVDAO.sol
```diff
@@ -3,25 +3,22 @@ pragma solidity 0.8.18;
...
     function updateNetworkFee(uint256 fee) external override {
-        uint64 previousFee = SSVStorageProtocol.load().networkFee;
-
-        SSVStorageProtocol.load().updateNetworkFee(fee);
+        StorageProtocol storage sp = SSVStorageProtocol.load();
+        uint64 previousFee = sp.networkFee;
 
-        emit NetworkFeeUpdated(previousFee, fee);
+        sp.updateNetworkFee(fee);
+        emit NetworkFeeUpdated(previousFee.expand(), fee);
```
andrew-blox, line 21: Make sure we have good unit test for this as we should not miss any expand / shrinks anywhere

I dont see any test edited to accompany this change which seems like we dont have a test for it.

andrew-blox (changes requested):

contracts/modules/SSVDAO.sol
mtabasco: Thanks, Andrew. I pushed the test. It confirms that the previous version didn't emit the right `previousFee` value.

andrew-blox (approved):

lior-blox (approved):

PR #253, opened 2023-08-16
Fee abstraction PoC

mtabasco:


contracts/interfaces/ISSVNetworkCore.sol
```diff
@@ -28,6 +30,8 @@ interface ISSVNetworkCore {
         bool whitelisted;
         /// @dev The state snapshot of the operator
         Snapshot snapshot;
```
mtabasco, line 32: explore to include it with the operator key

PR #256, opened 2023-08-24
A cluster is not liquidatable if it has zero validators

mtabasco:
A cluster is not liquidatable if it has zero validators, regardless of the settlement condition it meets (collateral/threshold). This keeps the cluster balance intact and its status active.


test/liquidate/liquidated-cluster.ts
```diff
@@ -123,4 +123,46 @@ describe('Liquidate Tests', () => {
     await utils.progressBlocks(2);
     expect(await ssvViews.getBalance(helpers.DB.owners[1].address, clusterEventData.operatorIds, clusterEventData.cluster)).to.be.equals(minDepositAmount * 2 - burnPerBlock * 2);
   });
+
+  it('Remove validator -> withdraw -> liquidate', async () => {
```
andrew-blox, line 127: I dont understand how this test passes with error saying ClusterNotLiquidatable (based on my commecnt above it should be an error), im assuming its getting this error not because of 0 validators but rather balance is not low enough.

andrew-blox (changes requested):

contracts/libraries/ClusterLib.sol
```diff
@@ -24,11 +24,15 @@ library ClusterLib {
...
+    ) internal pure returns (bool liquidatable) {
+        if (cluster.validatorCount != 0) {
+            if (cluster.balance < minimumLiquidationCollateral.expand()) return true;
+            uint64 liquidationThreshold = minimumBlocksBeforeLiquidation *
+                (burnRate + networkFee) *
+                cluster.validatorCount;
 
-        return cluster.balance < liquidationThreshold.expand();
+            return cluster.balance < liquidationThreshold.expand();
+        }
```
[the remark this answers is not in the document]
mtabasco: As I define the name of the return parameter `bool liquidatable`, it has the default value of `false` always.
So if the cluster has 0 validators, the function will always return `false`.
I checked gas usage with a more clear syntax like 
`if (cluster.validatorCount == 0) return false`
but its little more expensive.

test/liquidate/liquidated-cluster.ts
mtabasco: Yes, let me change the title of the test to be more clear.
In the last step, when trying to liquidate, the cluster has this state:
Balance: 157000000
Validators: 0

the minimum Liquidation Collateral is 200000000, greater than the cluster balance so it could be liquidatable, but with the current condition, as it has zero validators, it is not liquidated.

contracts/libraries/ClusterLib.sol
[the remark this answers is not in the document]
andrew-blox: @mtabasco understood, thats a cool hack

test/liquidate/liquidated-cluster.ts
andrew-blox: understood i just thought that the other code should for sure fail so i guessed maybe you had a flase positive going but awesome!

andrew-blox (approved):

lior-blox (approved):

PR #257, opened 2023-08-25
Remove registerAuth implementation

mtabasco:
The restriction to register operators/validators to whitelisted addresses is removed.
Important: The `RegisterAuth` library was removed from the contracts, but if it is linked again without changing the storage pointer, all previous authorized addresses (`RegisterAuth.authorization` mapping) will be active again.


contracts/deprecated/RegisterAuth.sol
```diff
@@ -6,8 +6,10 @@ struct Authorization {
     bool registerValidator;
 }
 
+/// @notice Deprecated. Notice that if the library is used again,
+/// all the values in AuthData.authorization will still be available.
 library RegisterAuth {
-    uint256 constant private SSV_STORAGE_POSITION = uint256(keccak256("ssv.network.storage.auth")) - 1;
+    uint256 private constant SSV_STORAGE_POSITION = uint256(keccak256("ssv.network.storage.auth")) - 1;
```
andrew-blox, line 12: General question (not related to this code)

We are not deleting the struct for registerAuth right?  I just wanna make sure as this storage is used and we need to make sure we dont use same name again, or anything that can corrupt storge etc ...

andrew-blox (changes requested):

test/deployment/deploy.ts
```diff
@@ -95,9 +95,9 @@ describe('Deployment tests', () => {
         expect(await ssvNetworkViews.getNetworkFee()).to.be.equals(0);
     });
 
-    it('Remove registerAuth from SSVNetwork contract', async () => {
+    it.skip('Remove registerAuth from SSVNetwork contract', async () => {
```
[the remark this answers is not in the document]
mtabasco: removed👍🏻

contracts/deprecated/RegisterAuth.sol
mtabasco: Yes, thats the reason I moved the library to the _deprecated_ folder instead of deleting it.

andrew-blox:
@lior-blox Please check this implementation is as you wish, i maybe dont understand 100% whats here would like a good double check here.

contracts/SSVNetwork.sol
```diff
@@ -12,7 +12,6 @@ import "./libraries/Types.sol";
 import "./libraries/CoreLib.sol";
```
lior-blox, line 12: lets completely remove all register auth ref including tests and struct

PR #262, opened 2023-09-19
Fix network withdrawals

mtabasco:
Update the DAO index when making a network withdrawal.

### DIFF
--- a/test/dao/network-fee-withdraw.ts
+++ b/test/dao/network-fee-withdraw.ts
@@ -68,4 +68,14 @@ describe('DAO Network Fee Withdraw Tests', () => {
     await expect(ssvNetworkContract.connect(helpers.DB.owners[3]).withdrawNetworkEarnings(amount
     )).to.be.revertedWith('Ownable: caller is not the owner');
   });
+
+  it('Withdraw network earnings sequentially when not enough balance reverts "InsufficientBalance"', async () => {
+    const amount = await ssvViews.getNetworkEarnings() / 2;
+
+    await ssvNetworkContract.withdrawNetworkEarnings(amount);
+    await ssvNetworkContract.withdrawNetworkEarnings(amount);
+
+    await expect(ssvNetworkContract.withdrawNetworkEarnings(amount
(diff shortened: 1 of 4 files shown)


test/dao/network-fee-withdraw.ts
```diff
@@ -68,4 +68,14 @@ describe('DAO Network Fee Withdraw Tests', () => {
...
+
+  it('Withdraw network earnings sequentially when not enough balance reverts "InsufficientBalance"', async () => {
+    const amount = await ssvViews.getNetworkEarnings() / 2;
+
+    await ssvNetworkContract.withdrawNetworkEarnings(amount);
+    await ssvNetworkContract.withdrawNetworkEarnings(amount);
+
+    await expect(ssvNetworkContract.withdrawNetworkEarnings(amount
+      )).to.be.revertedWithCustomError(ssvNetworkContract, 'InsufficientBalance');
+  });
```
andrew-blox, line 80: Please for these tests and all other tests where we do any withdraw lets add get assertions.
so in this particular test we do 3 withdrawls and one of them fails
After each one please do a getNetworkBalance and assert the number is correct.

Same for operator withdraw we should do a getOperatorEarnings and assert the number that its correct.

andrew-blox (changes requested):

test/dao/network-fee-withdraw.ts
```diff
@@ -68,4 +65,17 @@ describe('DAO Network Fee Withdraw Tests', () => {
     await expect(ssvNetworkContract.connect(helpers.DB.owners[3]).withdrawNetworkEarnings(amount
     )).to.be.revertedWith('Ownable: caller is not the owner');
   });
+
+  it('Withdraw network earnings sequentially when not enough balance reverts "InsufficientBalance"', async () => {
+    const amount = await ssvViews.getNetworkEarnings() / 2;
+
+    await ssvNetworkContract.withdrawNetworkEarnings(amount);
+    expect(await ssvViews.getNetworkEarnings()).to.be.equals(((networkFee * 13) + (networkFee * 11) - amount));
```
andrew-blox, line 73: I like the way you assert much more then the way i suggested :)

PR #263, opened 2023-09-22
A validator can voluntarily exit

mtabasco:
Allow a validator to start the exit procedure. Reference: [Link](https://github.com/bloxapp/SIPs/blob/main/sips/voluntary_exit.md)


contracts/modules/SSVClusters.sol
```diff
@@ -145,20 +146,7 @@ contract SSVClusters is ISSVClusters {
     ) external override {
         StorageData storage s = SSVStorage.load();
 
-        bytes32 hashedValidator = keccak256(abi.encodePacked(publicKey, msg.sender));
```
mtabasco, line 148: This part was refactored and moved into `ValidatorLib`

contracts/modules/SSVClusters.sol
```diff
@@ -344,4 +332,10 @@ contract SSVClusters is ISSVClusters {
 
         emit ClusterWithdrawn(msg.sender, operatorIds, amount, cluster);
     }
+
+    function exitValidator(bytes calldata publicKey, uint64[] calldata operatorIds) external override {
+        ValidatorLib.validateState(publicKey, operatorIds, SSVStorage.load());
+
+        emit ValidatorExited(publicKey, operatorIds);
+    }
```
andrew-blox, line 340: @lior-blox At what point do we minus the validator from the operator list?  If we dont, cant it be a hack that people can fill an operator with 500 validators then exit them and the operator is then trapped?

If its not dkg then how are you going to generate a keystore to perform the exit without a mnemonic?  Also how will it know the validators index (not beacon index but validator index from the mnemonic)

Also what if someone does this contract call to exit then does resharing with new operators with dkg right after this call?

test/validators/others.ts
```diff
@@ -0,0 +1,147 @@
...
+            firstCluster.operatorIds
+        )).to.be.revertedWithCustomError(ssvNetworkContract, 'ValidatorDoesNotExist');
+    });
+
+    it('Exiting a validator with the wrong operator setup reverts "IncorrectValidatorState"', async () => {
+        await expect(ssvNetworkContract.connect(helpers.DB.owners[1]).exitValidator(
+            helpers.DataGenerator.publicKey(1),
+            [4, 3, 2, 1]
+        )).to.be.revertedWithCustomError(ssvNetworkContract, 'IncorrectValidatorState');
+    });
```
andrew-blox, line 120: Please readjust the title to:
Exiting a validator with incorrect operators (unsorted list) reverts with "IncorrectValidatorState"

test/validators/others.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    it('Exiting a validator with the wrong operator setup reverts "IncorrectValidatorState"', async () => {
```
andrew-blox, line 122: Please readjust the title to:
Exiting a validator with incorrect operators (too many operators) reverts with "IncorrectValidatorState"

andrew-blox, line 122: Please add another test where you use operators 1,2,3,5 with title of:
Exiting a validator with incorrect operators (incorrect operator ids with valid amount) reverts with "IncorrectValidatorState"

andrew-blox (changes requested):

contracts/SSVNetwork.sol
```diff
@@ -216,6 +216,10 @@ contract SSVNetwork is
         _delegate(SSVStorage.load().ssvContracts[SSVModules.SSV_CLUSTERS]);
     }
 
+    function exitValidator(bytes calldata publicKey, uint64[] calldata operatorIds) external override {
```
andrew-blox, line 219: @lior-blox i believe we missing a param here of withdraw address thats wanted.  I dont think its good that we assume owner wants to receive to this address, especially with DKG, i assume with DKG that some validators will be created with some sort of multisig and maybe they dont want to withdraw there.

andrew-blox (changes requested):

contracts/SSVNetwork.sol
andrew-blox: This is irrelevant as we are not setting the withdraw address anyway.

andrew-blox (approved):

PR #270, opened 2023-10-30
Get Network total validators count

mtabasco:
Add views function `getNetworkValidatorsCount()` to retrieve the total active validators in the network.

### DIFF
--- a/contracts/modules/SSVViews.sol
+++ b/contracts/modules/SSVViews.sol
@@ -201,6 +201,10 @@ contract SSVViews is ISSVViews {
         return SSVStorageProtocol.load().validatorsPerOperatorLimit;
     }
 
+    function getNetworkValidatorsCount() external view override returns (uint32) {
+        return SSVStorageProtocol.load().daoValidatorCount;
+    }
+
     function getVersion() external pure override returns (string memory version) {
         return CoreLib.getVersion();
     }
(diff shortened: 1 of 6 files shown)


contracts/modules/SSVViews.sol
```diff
@@ -201,6 +201,10 @@ contract SSVViews is ISSVViews {
         return SSVStorageProtocol.load().validatorsPerOperatorLimit;
     }
 
+    function getNetworkValidatorsCount() external view override returns (uint32) {
+        return SSVStorageProtocol.load().daoValidatorCount;
```
andrew-blox, line 205: Just so we are aligned, this will not include liquidated validators

test/dao/operational.ts
```diff
@@ -24,4 +27,34 @@ describe('DAO operational Tests', () => {
         expect(await ssvNetworkContract.owner()).equals(helpers.DB.owners[4].address);
     });
 
+    it('Get the network validators count when updating it', async () => {
```
andrew-blox, line 30: let please add a scenario with a removed validator and a liquidated validator where we do a count after each

andrew-blox (changes requested):

contracts/modules/SSVViews.sol
mtabasco: correct, this does not include liquidated validators

andrew-blox (approved):

PR #277, opened 2023-11-10
[Update]: configuration for publishing npm package

Mohsen-T:
- Updated the package.json for publishing npm package
- Added .npmignore
- Added 'hardhat-abi-exporter' package for generating ABIs separately
- Ticket: [Publish NPM package with contracts](https://blox-ensemble.monday.com/boards/4802651069/pulses/4802659002)


hardhat.config.ts
```diff
@@ -1,73 +1,81 @@
...
+    gasPrice: 0.3,
+  },
+  abiExporter: {
+    path: './abis',
+    runOnCompile: true,
+    clear: true,
+    flat: true,
+    spacing: 2,
+    pretty: false,
+  },
```
mtabasco, line 78: Users will only interact with our main contracts SSVNetwork and SSVNetworkViews. So only these ABIs are needed. I think using the `only` property here you can got it.

mtabasco (changes requested):
Please add a reference for the publishing process in the docs.
Also, we need to setup the NPM_TOKEN in github. Assign it to me on the Monday task as a check to have it ready before merging this PR.
Is there a way of testing this publish process or we must wait until the next release? thanks!

andrew-blox (commented):
I see marco left a comment like this maybe its the same.  Why do we publish all the ABI's is it needed?  Can we do just ABI for network and views?

Just double checking this will be public package not private right?

docs/publish.md
```diff
@@ -0,0 +1,73 @@
...
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+```
+
+- Add your npm token `NPM_TOKEN` to the GitHub repository secrets (Settings > Secrets).
+
+## Publish Package
+
+- Push some changes to the `main` branch of the `ssv_network` GitHub repository.
```
mtabasco, line 72: Note that the publish action will occur when generating a release, not pushing to main.

docs/publish.md
```diff
@@ -0,0 +1,73 @@
...
+```bash
+npm login
+```
+
+- Enter your npm username, password, and email address when prompted.
+
+## Configure GitHub Actions for Automated Publishing
+
+- Create a `.github/workflows/publish.yaml` file in your project.
+- Define the npm publishing process using GitHub Actions:
```
mtabasco, line 38: @Mohsen-T I think instead of putting putting here the publish script, just reference to our `.github/workflows/publish.yaml` as example.

mtabasco (changes requested):

mtabasco:
> I see marco left a comment like this maybe its the same. Why do we publish all the ABI's is it needed? Can we do just ABI for network and views?
> 
> Just double checking this will be public package not private right?

@andrew-blox correct, public package

mtabasco (changes requested):
Minor changes

Mohsen-T:
> Please add a reference for the publishing process in the docs. Also, we need to setup the NPM_TOKEN in github. Assign it to me on the Monday task as a check to have it ready before merging this PR. Is there a way of testing this publish process or we must wait until the next release? thanks!

Introduced publish.md to the documentation and made updates to other Markdown files within the docs directory, enhancing the project's documentation quality.

docs/publish.md
Mohsen-T: okay

mtabasco (approved):

andrew-blox (approved):

PR #279, opened 2023-11-15
Holesky dev: change validatorsPerOperator limit to 565

mtabasco:
- Change validatorsPerOperator limit to 565
- Change version to v1.0.2


andrew-blox (approved):

Mohsen-T (approved):
"Change validatorsPerOperator limit to 565", I would need more context or access to the specific PR to precisely locate and address this change

PR #282, opened 2023-11-22
[Update]: deployment tests and update task and documentation

Mohsen-T:
- **Deployment and Verification on Goerli Testnet:** Successfully deployed all contracts on the Goerli testnet and subsequently verified each one post-deployment.
- **Contract Upgrade and Testing:** Upgraded the SSVNetwork to SSVNetworkValidatorsPerOperator, followed by re-verification and testing of the new functionalities.
- **Module Update Testing:** Conducted comprehensive testing to ensure all requirements for updating modules were met.
- **Documentation Updates:** Enhanced the /docs section with additional descriptions, focusing on contract verification, issue resolution, module updates, and related topics.
- **Ticket:** [Deployment tests](https://blox-ensemble.monday.com/boards/4802651069/pulses/5539718732?userId=51361533)


contracts/upgrades/stage/goerli/modules/SSVOperatorsV2.sol
```diff
@@ -0,0 +1,22 @@
...
+    using Types64 for uint64;
+    /****************************************/
+    /* Upgraded Operator External Functions */
+    /****************************************/
+
+    function getOperatorById(uint64 operatorId) external view returns (address, uint256, uint32, address, bool, bool) {
+        StorageData storage s = SSVStorage.load();
+        Operator memory operator = s.operators[operatorId];
+        address whitelisted = s.operatorsWhitelist[operatorId];
+        bool isPrivate = whitelisted == address(0) ? false : true;
```
andrew-blox, line 17: Just a solidity question (im curious not a CR) can you do this as well or its not valid in solidity:
bool isPrivate = !whitelisted == address(0)

andrew-blox: This file is just for testing in general right?

docs/local-dev.md
```diff
@@ -99,3 +124,14 @@ https://goerli.etherscan.io/address/0x227...#code
 ```
 
 After this action, you can go to the proxy contract in Etherscan and start interacting with it.
+
+### How to resolve issues during the verification
+
+- Error: no such file ro directory, open ‘…/artifacts/build-info/XXXX...XXXX.json’
```
andrew-blox, line 130: ro to or
If copy and pasted then i guess also needs to be fixed in code

andrew-blox (changes requested):

tasks/deploy.ts
```diff
@@ -76,48 +70,47 @@ The deployer account used will be the first one returned by ethers.getSigners().
...
-        const ssvTokenFactory = await ethers.getContractFactory('SSVToken');
-        const ssvToken = await ssvTokenFactory.deploy();
-        await ssvToken.deployed();
-
-        console.log(`SSV Network Token deployed to: ${ssvToken.address}`);
-
-    });
+subtask('deploy:module', 'Deploys a new module contract')
+  .addParam('module', 'SSV Module', null, types.string)
+  .addParam('contract', 'SSV Module Contract', null, types.string)
```
mtabasco, line 75: @Mohsen-T I understand the level of flexibility you want to provide here, but I think this overcomplicates the process of updating modules. Generally, when updating modules for all cases from bugs to new features, we will keep the same version over all envs, from stage (dev) to mainnet.
The reason to use a different contract name when upgrading SSVNetwork or SSVNetworkViews is that for some environments, we need to update state variables via `initializer` function, but only in one environment for testing purposes.
I'm open to hearing a reason to keep the `contract` parameter here, if not, please remove it. thanks!

mtabasco (changes requested):
As this is a test task, @Mohsen-T please remove the deployment metadata update and the SSVOperatorsV2 test contract. If you remove the `contract` parameter in `update:module` task, I think there is only changes in the docs. thanks!

contracts/upgrades/stage/goerli/modules/SSVOperatorsV2.sol
Mohsen-T: Yes, only for testing

Mohsen-T: The two lines of code you've provided are syntactically correct and will work as expected in Solidity

tasks/deploy.ts
Mohsen-T: removed SSVStorageV2.sol and contract parameter in update module script

andrew-blox (approved):

mtabasco (approved):

PR #284, opened 2023-11-27
[Review] Balance calculation when operators will be removed

Mohsen-T:
- Reviewed all the code and test scripts related to the SSVOperators.remoteOperator() function
- Confidently affirm that there are no errors or logical inconsistencies in the code
- Could you please clarify the rationale for using the Operator's whitelisted variable independently, as opposed to utilizing the SSVStorage's StorageData structure's operatorsWhitelist mapping variable?
- Ticket: [Review/fix balance calculations when removing operators](https://blox-ensemble.monday.com/boards/4802651069/pulses/5280973671)


contracts/modules/SSVOperators.sol
```diff
@@ -62,6 +62,7 @@ contract SSVOperators is ISSVOperators {
         operator.snapshot.balance = 0;
         operator.validatorCount = 0;
         operator.fee = 0;
+        operator.whitelisted = false;
```
andrew-blox, line 65: Just so i understand it all, this is the main thing in the PR right?
And the only implication of this was that in the getOperatorById view command it would be inaccurate for whitelisted true but removed operator right?

mtabasco (approved):

test/sanity/balances.ts
```diff
@@ -55,7 +55,7 @@ describe('Balance Tests', () => {
     }
   });
 
-  it('Check cluster balance after removing operator, progress blocks and confirm', async () => {
+  it.only('Check cluster balance after removing operator, progress blocks and confirm', async () => {
```
mtabasco, line 58: please remove `only`

mtabasco (changes requested):

mtabasco (approved):

PR #286, opened 2023-12-01
[Update]: Optimize functions that operate with private operators

Mohsen-T:
- Removed a unnecessary code in the registerValidator()
- Ticket: [Optimize functions that operate with private operators](https://blox-ensemble.monday.com/boards/4802651069/pulses/5282998040?userId=51361533)


mtabasco:
Outdated because of multi-op-whitelist

PR #287, opened 2023-12-17
Bulk registration of validators

mtabasco:
- Add `bulkRegisterValidator` function
- Refactor `registerValidator` and `bulkRegisterValidator`


hardhat.config.ts
```diff
@@ -46,6 +46,7 @@ const config: HardhatUserConfig = {
     } as SSVNetworkConfig,
     hardhat: {
       allowUnlimitedContractSize: true,
+      gas: 5000000,
```
mtabasco, line 49: I had problems using hardhat network when bulk registering validators. With less than 10 validators I reached 30M gas, which is not the real cost. I set this param to fix it.

test/helpers/gas-usage.ts
```diff
@@ -65,21 +77,33 @@ const MAX_GAS_PER_GROUP: any = {
   [GasGroup.EXECUTE_OPERATOR_FEE]: 52000,
   [GasGroup.REDUCE_OPERATOR_FEE]: 51900,
 
-  [GasGroup.REGISTER_VALIDATOR_EXISTING_CLUSTER]: 202000,
+  [GasGroup.REGISTER_VALIDATOR_EXISTING_CLUSTER]: 201200,
```
mtabasco, line 80: @andrew-blox Refactoring the code gets savings on the gas costs, around 1500 gas per GasGroup, but as this version uses `ssv-keys` to generate the shares, all values are increased.

test/helpers/json/validatorKeys.json
```diff
@@ -18,5 +18,45 @@
     "id": 4,
     "privateKey": "0x52e4bdbdedd85e9e8a77c2f60d083a0693d6f7b2bd717f642a02514528db4278",
     "publicKey": "0x9150572051c3496a67207b4caa371dfba34f127318a7aef145ebdba6e0de506c292af31e20831b0c537ab7478508d3e9"
+  },
```
mtabasco, line 21: @Mohsen-T I had to add new keys to support bulk registration.

hardhat.config.ts
andrew-blox: What does this do? Why was reaching so high?  WIth this param that you added is the gas correct now?  

Not related but the param above it "allowUnlimitedContractSize" i think maybe its best we put that back to true since were diamond now?

Mohsen-T (approved):

hardhat.config.ts
mtabasco: Using default hardhat config, for some reason moving from bulk registering 8 validators (less than 1M gas) to 10 results in more than 30M gas. Specifying the gas in the network config results in real numbers close to holesky ones.
Agree on removing `allowUnlimitedContractSize`

PR #290, opened 2024-01-08
[Update]: Fuzz Test by Echidna

Mohsen-T:
- integrated the Echidna with hardhat project
- added echidna commands in the project configuration
- configured the env to run test on a forked network
- configured echidna.yaml for customized testing
- added fuzz testing for SSVOperators.sol, SSVDAO.sol, and SSVClusters.sol
- fixed 2 issues in the SSVOperators.sol


package.json
```diff
@@ -7,9 +7,12 @@
     "lint:fix": "eslint --fix . --ext .ts",
     "solidity-coverage": "NO_GAS_ENFORCE=1 npx hardhat coverage",
     "slither": "slither contracts --solc-remaps @openzeppelin=node_modules/@openzeppelin",
-    "size-contracts": "npx hardhat size-contracts"
+    "size-contracts": "npx hardhat size-contracts",
+    "install-echidna": "brew install echidna",
+    "echidna": "echidna --config echidna.yaml --contract"
```
mtabasco, line 12: This does not work for me, maybe it requires an extra parameter specifying the contract?

mtabasco (changes requested):

package.json
```diff
@@ -7,9 +7,12 @@
     "lint:fix": "eslint --fix . --ext .ts",
     "solidity-coverage": "NO_GAS_ENFORCE=1 npx hardhat coverage",
     "slither": "slither contracts --solc-remaps @openzeppelin=node_modules/@openzeppelin",
-    "size-contracts": "npx hardhat size-contracts"
+    "size-contracts": "npx hardhat size-contracts",
+    "install-echidna": "brew install echidna",
```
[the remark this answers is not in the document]
Mohsen-T: added description for installation of Echidna in the README

package.json
Mohsen-T: found out issue and fixed

.echidna.test.js
```diff
@@ -0,0 +1,26 @@
+const { execSync } = require('child_process');
+const chalk = require('chalk');
```
mtabasco, line 2: I think it's enough with monochromic logs🙂. No need to add an extra dependency.

.echidna.test.js
```diff
@@ -0,0 +1,26 @@
...
+const contract = process.argv[2];
+
+const exec = commands => {
+  try {
+    execSync(commands, { stdio: 'inherit', shell: true }); // Specify the shell if needed
+  } catch (error) {}
+};
+
+const echidnaPath = '/usr/local/bin/echidna';
```
mtabasco, line 11: This does not work for me. I think the command can be formed using `echidna` command instead of the path, and let the user set it accessible via `$PATH` env var

contracts/modules/SSVDAO.sol
```diff
@@ -15,6 +15,8 @@ contract SSVDAO is ISSVDAO {
     uint64 private constant MINIMAL_LIQUIDATION_THRESHOLD = 100_800;
 
     function updateNetworkFee(uint256 fee) external override {
+        if (fee == 0) revert FeeTooLow();
```
mtabasco, line 18: I don't think we must restrict the networkFee to be greater than 0.
Maybe the DAO wants it temporarily for some reason. @andrew-blox thoughts?

contracts/modules/Clusters.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity 0.8.18;
```
mtabasco, line 2: not related to this file but, can you move the test contracts to another folder? so we can skip them from generating the npm package, etc?
Also, prefix/suffix them to quickly see they are test files.

mtabasco (changes requested):

PR #294, opened 2024-02-20
[Update]: Fuzz test using the Echidna

Mohsen-T:
- Used `echidna.yaml` for the configuration of Echidna.
- Developed a script named `.echidna.test.js` for wrapping parameters like contract name, path, etc.
- Developed `Operators.sol` for fuzz testing of `SSVOperators.sol`, `Clusters.sol` for `SSVClusters.sol`, and `DAO.sol` for `SSVDAO.sol`.
- Found some issues from the fuzz testing; I would like to discuss them in the presentation.
- Used the `assertion` mode for fuzz testing, among options such as `property`, `assertion`, `dapptest`, etc.


Stam098c:
> * Used `echidna.yaml` for the configuration of Echidna.
> * Developed a script named `.echidna.test.js` for wrapping parameters like contract name, path, etc.
> * Developed `Operators.sol` for fuzz testing of `SSVOperators.sol`, `Clusters.sol` for `SSVClusters.sol`, and `DAO.sol` for `SSVDAO.sol`.
> * Found some issues from the fuzz testing; I would like to discuss them in the presentation.
> * Used the `assertion` mode for fuzz testing, among options such as `property`, `assertion`, `dapptest`, etc.

PR #307, opened 2025-02-27
Upgrade contract to increase max validators

mtabasco:


contracts/upgrades/mainnet/SSVNetworkUpgradeValidatorsPerOperator.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity 0.8.24;
+
+import "../../SSVNetwork.sol";
+
+contract SSVNetworkUpgradeValidatorsPerOperator is SSVNetwork {
+    function initializev2(uint32 validatorsPerOperatorLimit_) external reinitializer(_getInitializedVersion() + 1) {
```
riccardo-ssvlabs:
```suggestion
    function initializev2(uint32 validatorsPerOperatorLimit_) external reinitializer(2) {
```
I think this should do the job.

This means it can be called only once but probably it's still safer to add onlyOwner.

mtabasco:
Changed `initializev2` auth to `onlyOwner`, removing initializer logic.
tested on Sepolia:
SSV Network: https://sepolia.etherscan.io/address/0x132f3052cB781c95AFb9a5e2C4Be3678a94219DE#writeProxyContract
SSV Views: https://sepolia.etherscan.io/address/0xAfdb141Dd99b5a101065f40e3D7636262dce65b3#readProxyContract

riccardo-ssvlabs (dismissed):
Approving with the following discussed assumptions:

* only the owner will be able to call the new `initializev2` function
* we are ok with having this function callable multiple times, so it is not a real initializer anymore but it is more similar to a set function.

mtabasco:
Closed in favor of v2.0.0 release

PR #321, opened 2025-12-02
Feat/eth migration

venimir-ssv:


contracts/modules/SSVClusters.sol
```diff
@@ -357,4 +417,57 @@ contract SSVClusters is ISSVClusters {
...
+        // compute cluster data using ETH fields
+        (uint64 clusterIndex, uint64 burnRate) = OperatorLib.updateClusterOperators(
+            operatorIds,
+            true,
+            cluster.validatorCount,
+            s,
+            sp
+        );
+
+        cluster.balance += msg.value;
```
mtabasco: @venimir-ssv this should be `cluster.balance = msg.value;` to reset to balance to the value of ETH sent, right?

contracts/modules/SSVClusters.sol
```diff
@@ -357,4 +417,57 @@ contract SSVClusters is ISSVClusters {
...
+        for (uint256 i; i < opIds.length; ++i) {
+            ISSVNetworkCore.Operator storage operator = s.operators[opIds[i]];
+            operator.ensureETHDefaults();
+        }
+
+        // compute cluster data using ETH fields
+        (uint64 clusterIndex, uint64 burnRate) = OperatorLib.updateClusterOperators(
+            operatorIds,
+            true,
+            cluster.validatorCount,
```
mtabasco, line 435: @venimir-ssv We should remove the current `cluster.validatorCount` amount from operators validatorCount (SSV). I see you add the `cluster.validatorCount` to `operator.ethValidatorCount` in `OperatorLib.updateClusterOperators`, but please double check we remove the amount from the operators SSV validatorCount

contracts/modules/SSVClusters.sol
```diff
@@ -357,4 +417,57 @@ contract SSVClusters is ISSVClusters {
...
+        cluster.validateClusterIsNotLiquidated();
+
+        uint256 ssvBalance = cluster.balance;
+        // migrate operators to ETH defaults if needed
+        uint64[] memory opIds = operatorIds;
+        for (uint256 i; i < opIds.length; ++i) {
+            ISSVNetworkCore.Operator storage operator = s.operators[opIds[i]];
+            operator.ensureETHDefaults();
+        }
+
```
mtabasco, line 430: @venimir-ssv before `updateClusterOperators` to add ETH accounting, we should settle SSV earnings for operators. I think a call to `updateSnapshotStSVV` will work.

mtabasco (changes requested):
migrateClusterToETH

contracts/modules/SSVClusters.sol
venimir-ssv: @mtabasco if we remove the cluster.validatorCount, I think this will break the backward compatibility hashClusterData will produce different hashes

contracts/modules/SSVClusters.sol
venimir-ssv: Yes great catch thanks

contracts/modules/SSVClusters.sol
venimir-ssv: Will refactor

venimir-ssv: fixed here: https://github.com/ssvlabs/ssv-network/pull/321/commits/44371027d4cff5eb43d4bcc43e730efbe6cc9d73

contracts/modules/SSVClusters.sol
venimir-ssv: fixed here: https://github.com/ssvlabs/ssv-network/pull/321/commits/44371027d4cff5eb43d4bcc43e730efbe6cc9d73

mtabasco:
@venimir-ssv I think we can close this as we work with feat/eth-eb-merge

mtabasco:
closed because it was integrated in https://github.com/ssvlabs/ssv-network/tree/feat/eth-eb-merge

PR #322, opened 2025-12-03
Support validator EB increase

mtabasco:
This PR introduces support for the Effective Balance (EB) oracle system, which updates the SSV Network's accounting model from validator-count-based fees to EB-weighted fees (vUnits). This allows operator and network fees to scale proportionally with actual staked ETH rather than treating all validators equally.

Specs: https://hackmd.io/@mtabasco/EB-submission-contracts

**Backwards Compatibility**
- Before oracles activate, all EB storage is zero
- `getVUnits` falls back to `validatorCount` * `VUNITS_PRECISION`
- Operator snapshots fall back when `operatorVUnits` == 0
- No migration required; EB accounting activates per-cluster on first `updateClusterBalance`

**Known Issues**
Stack-too-deep in `updateClusterBalance`

### DIFF
--- a/contracts/SSVNetwork.sol
+++ b/contracts/SSVNetwork.sol
@@ -271,6 +271,17 @@ contract SSVNetwork is
         _delegate(SSVStorage.load().ssvContracts[SSVModules.SSV_CLUSTERS]);
     }
 
+    function updateClusterBalance(
+        uint64 blockNum,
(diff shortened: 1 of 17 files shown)


mtabasco:
@yurii-ssv I think we can close this as we work with feat/eth-eb-merge

PR #327, opened 2025-12-11
refactor: centralize validator register/remove flows in SSVClusters

venimir-ssv:


yurii-ssv (approved):

contracts/modules/SSVClusters.sol
```diff
@@ -850,21 +699,17 @@ contract SSVClusters is ISSVClusters {
...
-    function _performLiquidation(
+    function _executeLiquidation(
         address clusterOwner,
+        address liquidator,
         bytes32 clusterId,
         uint64[] calldata operatorIds,
         Cluster memory cluster,
-        uint8 version
+        uint8 version,
+        bool payLiquidator
```
mtabasco: @venimir-ssv this is not used.

contracts/modules/SSVClusters.sol
```diff
@@ -850,21 +699,17 @@ contract SSVClusters is ISSVClusters {
...
         if (!liq) return;
 
-        _performLiquidation(
-            clusterOwner,
-            clusterId,
-            operatorIds,
-            cluster,
-            version
-        );
+        _executeLiquidation(clusterOwner, clusterOwner, clusterId, operatorIds, cluster, version, false);
```
mtabasco: @venimir-ssv here, the 2nd parameter should be msg.sender, as the reward will always go to the caller. I think this was not introduced by you, but please change it 🙏🏻

mtabasco (changes requested):
small changes

contracts/modules/SSVClusters.sol
venimir-ssv: https://github.com/ssvlabs/ssv-network/pull/327/commits/c070788a36d70e8b95e183de13c5db540e0a0299

fixed

contracts/modules/SSVClusters.sol
venimir-ssv: https://github.com/ssvlabs/ssv-network/pull/327/commits/7ed91e7d24de7045d6a6e487fba3c3fa9707f281

changed

PR #328, opened 2025-12-11
Migrate to hardhat v3

yurii-ssv:


scripts/deploy-ssv-network.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const operatorsModAddress = parseArg("operators-mod");
+  const clustersModAddress = parseArg("clusters-mod");
+  const daoModAddress = parseArg("dao-mod");
+  const viewsModAddress = parseArg("views-mod");
+  const ssvTokenAddress = parseArg("ssv-token");
+
+  console.log(`Deploying SSVNetwork proxy on ${targetNetwork}`);
+
+  const { address: implAddress } = await deployContract(ethers, "SSSVNetwork");
```
mtabasco: typo: SSVNetwork

mtabasco (changes requested):
Good job Yurii!
My review:

- Because the openzeppelin upgrades plugin was removed, we need to be extra cautious when upgrading SSVNetwork/SSVNetworkViews, but following the deploy implementation -> call to `upgradeTo`pattern is OK. I think we should use a storage layout tool (HH plugin, 3rd party tool...) to take a snapshot of the current storage and attach it to a deployment/tag. Then, not only locally but on the CI, we can check the compatibility with new changes taking a new snapshot and comparing. Not for this PR, or course.
- In `deploy-all.ts`, it requires `ssvToken` to exist in the hh `hardhat.config`. This is different from the previous HHv2, where if it's a `-development` network, the script checks if the token exists in `hardhat.config` and if not, it deploys a mock token. I think the new approach is solid. we can create a new hoodi-dev network in `hardhat.config`, deploy the token and put it there.
- Can these 3 be integrated into one `deploy-module` script with a flag `attach` which defaults to `false`? or do you think there could be moments where we deploy a module, wait for something and later attach?
> deploy-module.ts (deploy only)
> attach-module.ts (attach only)
> update-module.ts (deploy + attach, always)

- typo: SSSVNetwork
- `just` usage. As all the scripts in `Justfile` are one-liners, I think the use of `Just` here does not add much value comparing to add the scripts to `package.json`, but adds an extra tool usage. Unless you see the extra value, please use `package.json`.
- Deployment tracking: This is great! It would be great to add metadata info á la SSVLM/Based Apps, as it's useful for external teams.
```
 {
    "chainId": 560048,
    "deploymentBlock": 260728
  },
  "deploymentParams": {
    "feeExpireTime": 3600,
    "feeTimelockPeriod": 86400,
...
}
```
- ABI loader, contract sizer: why not use the abi-exporter and contract-sizer plugins? some are from time ago but can do the job well. I'm not sure if dealing with _buildInfo_ files can lead us to errors.

scripts/deploy-ssv-network.ts
yurii-ssv: Opezeppelin's plugin is unfortunately still not supported on hh v3, but they are working on it already: https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1191
I think we can wait for now and return to the upgradeability checks later (maybe plugin would be adapted already)

yurii-ssv: I think we should keep module deploying scripts separated. It will allow us to have more flexibility and handle different scenarios especially while still developing and doing redeployments/upgrades daily

yurii-ssv: Abi loader and contract sizer are also stuck with hardhat v2, adding them created a lot of dependency tree issues

yurii-ssv: And I want to keep Justfile, it also allows us to be more flexible and centralize all repo interactions. I think we should keep updating recipes to auto verify after deployments, load abis or maybe update update some configs/call functions

mtabasco: Justfile: ok, let's keep it for now in this dev cycle with SSV Staking, you are right that this give us more room to create more sophisticated scripts.

PR #332, opened 2025-12-16
Allow migrating liquidated SSV clusters without double-counting opera…

venimir-ssv:
…tors

### DIFF
--- a/contracts/libraries/OperatorLib.sol
+++ b/contracts/libraries/OperatorLib.sol
@@ -6,7 +6,6 @@ import {ISSVWhitelistingContract} from "../interfaces/external/ISSVWhitelistingC
 import {StorageData} from "./SSVStorage.sol";
 import {StorageProtocol} from "./SSVStorageProtocol.sol";
 import {Types64, Types256} from "./Types.sol";
-import "./CoreLib.sol";
 import "./SSVStorageEB.sol";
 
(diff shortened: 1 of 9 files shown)


contracts/libraries/OperatorLib.sol
```diff
@@ -220,9 +211,9 @@ library OperatorLib {
 
             ISSVNetworkCore.Operator storage operator = s.operators[operatorId];
 
-            if (operator.version != CoreLib.VERSION_ETH) {
+            if (operator.ethSnapshot.block == 0) {
                 updateSnapshotStSSV(operator, operatorId);
-                if (increaseValidatorCount) {
+                if (increaseValidatorCount && !isClusterLiquidated) {
```
mtabasco, line 216: @venimir-ssv I don't understand this one. Why check if the cluster is not liquidated there? I think we should be cluster/validator agnostic in this library and put only operator-related logic.

mtabasco (approved):

contracts/libraries/OperatorLib.sol
venimir-ssv: We only need that guard during SSV→ETH migration of a liquidated cluster. Liquidation already decremented operator.validatorCount on the SSV side; when we initialize ETH snapshots, updateClusterOperators(..., increaseValidatorCount=true) would otherwise subtract the same SSV count again. The flag tells the operator helper to skip that one-time SSV decrement.

PR #340, opened 2025-12-19
Staking contract

mtabasco:
Reference: https://hackmd.io/@mtabasco/ssv_staking


contracts/modules/SSVStaking.sol
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        // 4. Mint cSSV receipt tokens 1:1
+        ICSSVToken(cssv).mint(msg.sender, amount);
+
+        emit Staked(msg.sender, amount);
+    }
+
+    function requestUnstake(uint256 amount) external {
+        if (amount == 0) revert ZeroAmount();
```
venimir-ssv, line 50: Shall we check if there is pending unstake process already active?

contracts/libraries/SSVStorageStaking.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity 0.8.24;
+
+struct UnstakeRequest {
+    /// @notice Amount of cSSV burned and pending to be withdrawn as SSV
+    uint192 amount;
+    /// @notice Timestamp after which the pending unstake can be withdrawn
+    uint64 unlockTime;
```
venimir-ssv, line 8: Shall we use block number instead of timestamp?

contracts/token/CSSVToken.sol
```diff
@@ -0,0 +1,40 @@
...
+pragma solidity 0.8.24;
+
+import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+interface ISSVStaking {
+    function onCSSVTransfer(address from, address to) external;
+}
+
+contract CSSVToken is ERC20 {
+    error NotSSVStaking();
```
venimir-ssv, line 11: I think it's better to move the errors in ISSVStaking

venimir-ssv (approved):

contracts/modules/SSVViews.sol
```diff
@@ -463,6 +467,63 @@ contract SSVViews is ISSVViews {
         return SSVStorageProtocol.load().ethDaoValidatorCount;
     }
 
+    function cooldownDuration() external pure override returns (uint256) {
+        return 7 days; // TODO get the stored value
+    }
+
+    function totalStaked() external view override returns (uint256) {
+        address cssv = SSVStorageStaking.load().cssv;
+        return cssv == address(0) ? 0 : ICSSVToken(cssv).totalSupply();
```
yurii-ssv: maybe it will be better to enforce saving cssv address as const/immutable in constructor/initializer to always be sure that it is not address(0)

contracts/modules/SSVStaking.sol
mtabasco: yes, it's done here:
`if (s.withdrawals[msg.sender].amount != 0) revert CooldownActive();`

contracts/token/CSSVToken.sol
mtabasco: As this will be an isolated contract, I think it is good to not have interface dependencies with SSV Network contracts.

yurii-ssv (approved):

contracts/modules/SSVViews.sol
mtabasco: Yes, cSSV should be set when we upgrade the contract, so this check for address(0) can be removed. let me do it

contracts/libraries/SSVStorageStaking.sol
mtabasco: let's keep timestamp as we use it on declaring new operator fees, etc

PR #343, opened 2025-12-22
Set default quorumBps on deployment and update docs

venimir-ssv:


scripts/deploy-all.ts
```diff
@@ -6,6 +6,11 @@ async function main() {
   const targetNetwork = parseArg("network");
   const ethers = await getEthers(targetNetwork);
   const deployer = await getDeployer(ethers);
+  const quorumBps = Number(process.env.QUORUM_BPS ?? "6700");
```
yurii-ssv: maybe better to throw an error here if nothing is configured in env instead of falling back to hardcoded default?

scripts/deploy-ssv-network.ts
```diff
@@ -11,6 +11,11 @@ async function main() {
   const daoModAddress = parseArg("dao-mod");
   const viewsModAddress = parseArg("views-mod");
   const ssvTokenAddress = parseArg("ssv-token");
+  const quorumBps = Number(process.env.QUORUM_BPS ?? "6700");
```
yurii-ssv: same here

yurii-ssv (approved):

scripts/deploy-all.ts
venimir-ssv: okay will fix this

venimir-ssv: https://github.com/ssvlabs/ssv-network/pull/343/commits/525a4f96a3c529bbda538f22a02836cd41adce35

scripts/deploy-ssv-network.ts
venimir-ssv: https://github.com/ssvlabs/ssv-network/pull/343/commits/525a4f96a3c529bbda538f22a02836cd41adce35

PR #349, opened 2025-12-24
register tests added

venimir-ssv:


test/unit/SSVClusters/helpers/clusterHelpers.ts
```diff
@@ -0,0 +1,64 @@
+import { ClusterStruct } from "../types/cluster.js";
```
yurii-ssv, line 1: i think its better to define Cluster somewhere in our TS types and have shared types file for all tests

test/unit/SSVClusters/helpers/clusterHelpers.ts
```diff
@@ -0,0 +1,64 @@
...
+      }
+    } catch (error) {
+      // If it's our error, rethrow it
+      if (error instanceof Error && error.message.includes("unexpectedly emitted")) {
+        throw error;
+      }
+      // skip non-matching logs
+    }
+  }
+}
```
yurii-ssv, line 64: `mustEmitEvent` and `mustNotEmitEvent` look redundant, we can use
`expect(tx).to.emit(event).withArgs(args)` and `expect(tx).to.not.emit(event)`

test/unit/SSVClusters/helpers/clusterHelpers.ts
yurii-ssv: same for helpers, I think better to have shared `helpers` across all tests

test/unit/SSVClusters/register.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    const tx = await network.registerValidator(
+      publicKey,
+      operatorIds,
+      DEFAULT_SHARES,
+      0,
+      { ...EMPTY_CLUSTER },
+      { value: deposit }
+    );
+
+    await expect(tx).to.emit(network, "ValidatorAdded");
```
yurii-ssv, line 48: I think its better to move all event/error names to a separated file as constants (like
    `const VALIDATOR_ADDED_EVENT_NAME = "ValidatorAdded"`
) instead of hardcoding, it will allow to have shared constants across all test files and in case of any changes we will have to only update the constant

yurii-ssv: same for other values, I think we should go with the approach where we have

./test/helpers/constants.ts
./test/helpers/events.ts
./test/helpers/errors.ts
...

everything that can be shared (default cluster, default keys, ...) should be shared via constants, everything specific to the exact file - declared as const on the file level

test/unit/SSVClusters/register.test.ts
```diff
@@ -0,0 +1,184 @@
...
+      call: (ctx: any) =>
+        ctx.network.registerValidator(
+          "0x1234",
+          ctx.operatorIds,
+          DEFAULT_SHARES,
+          0,
+          { ...EMPTY_CLUSTER }
+        ),
+    },
+  ];
```
yurii-ssv, line 109: I think we should also add the possibility to check params (if custom error has anything inside we should check it with `revertedWithCustomError().withArgs()`)

test/unit/SSVClusters/helpers/clusterHelpers.ts
```diff
@@ -0,0 +1,64 @@
...
+      .connect(owner)
+      .registerOperator(makeOperatorKey(i + 1), 0, false);
+    await tx.wait();
+    operatorIds.push(i + 1);
+  }
+
+  return operatorIds;
+}
+
+export function asClusterStruct(cluster: any): ClusterStruct {
```
yurii-ssv, line 25: I think this can be avoided, if events would be checked with `expect().to.emit()`, all contract calls can explicitly expect Cluster type so we will not need to cast types

test/unit/SSVClusters/helpers/clusterHelpers.ts
venimir-ssv: Agree

PR #351, opened 2026-01-05
Feature/hotfixes and tests

yurii-ssv:


andrew-blox:
in _bulkRegisterValidator you have 
        for (uint i; i < validatorsLength; ++i) {
            ValidatorLib.registerPublicKey(publicKeys[i], operatorIds, owner, s);
        }
        bytes32 hashedCluster = cluster.validateClusterOnRegistration(owner, operatorIds, s);

This means that if i bulk registered 80 then failed on cluster data it will be super expensive revert, suggesting to validate cluster data before bulk registration of pubkeys

andrew-blox:
in registerPublicKey we save validators like this
s.validatorPKs[hashedPk] = bytes32(uint256(keccak256(abi.encodePacked(operatorIds))) | uint256(0x01));

What im suggesting is huge change and prob not backward compatible but to me this is a waste to just store 1 or true whatever.  Like if were gonna take this storage we might as well put something useful like the operators its using.  If we did that then when i do getValidator i will get not just true but the actual operators this account registered this validator to.

Remember if we registered the same validator twice with different owners it will be a different hash to there will still be multiple entries for matching validators

andrew-blox:
in validateClusterOnRegistration you have 

        // todo owner can override ssv cluster here, refactor this check
        if (clusterData == bytes32(0) && clusterDataSSV!= bytes32(0)) {
            revert ISSVNetworkCore.IncorrectClusterVersion();
        }

I can technically migrate without calling migrate function.  I can liquidate myself / remove validators (clusterData == bytes32(0)) then register a validator and i will migrate.  BUt since i didnt officially migrate im not sure we will pay the operators correctly i would assume we pay the operators ssv but with amounts of .00x as in the accounting would pay eth balances but in ssv.

This needs to be reviewed to make sure its true, tight unit testing here would be great

andrew-blox:
function ensureETHDefaults(ISSVNetworkCore.Operator storage operator) internal {
        if (operator.ethSnapshot.block == 0) {
            operator.ethSnapshot = ISSVNetworkCore.Snapshot({block: uint32(block.number), index: 0, balance: 0});
        }
        if (operator.ethFee == 0 && operator.fee != 0) {
            operator.ethFee = defaultOperatorEthFee();
        }
    }

what is operator.fee?  Is it ssv fee?  If so then an operator that has 0 fee in ssv wont fall here and give a default operator fee

andrew-blox:
i see we have isLiquidatable and isLiquidatableWithEB
when registering a validator we call isLiquidatable should we instead be calling isLiquidatableWithEB?
we call isLiquidatable as of now in the register validator funct (updateClusterOnRegistration)

Also isLiquidatable get call should prob call isLiquidatableWithEB internally
I didnt check but also reactivate and withdraw should be affected and migrate

andrew-blox:
in updateClusterOperators we have 
                if (increaseValidatorCount && !isClusterLiquidated) {
                    operator.validatorCount -= deltaValidatorCount;
                }

shouldnt it be += as increaseValidatorCount = true??

andrew-blox:
in _updateOperatorVUnits you have 
if (deltaPositive) seb.operatorEthVUnits[operatorId] += deltaAbs;
                        else seb.operatorEthVUnits[operatorId] -= deltaAbs

im not 100% understanding but just want to make sure deltaAbs can never be higher than seb.operatorEthVUnits[operatorId] right?

andrew-blox:
in getBalanceSSV i see you have 
cluster.updateBalanceWithEB(hashedCluster, clusterIndex, sp.currentNetworkFeeIndexSSV());

shouldnt it be cluster.updateBalance?  Lots of other places have this logic too i guess just make sure they also call updateBalance instead of updateBalanceWithEB as i didnt check all places that should have this

andrew-blox:
I think maybe theres some very edge case race condition, in updateBalanceWithEB you have 
uint64 vUnits = getVUnits(clusterId, cluster.validatorCount);

Now if the oracle calls to update some cluster by accident calls twice same update func on same cluster really fast in same block, i dont know if the getVunits will give old or updated vunits and how will it apply to the balance.

Maybe theres nothing wrong here i think you guys should verify though

andrew-blox:
migrateClusterToETH has 2 major issues first is already stated before using isLiquidatable instead of the eth one

Second is you get the burn rate at the beginning
(uint64 clusterIndex, uint64 burnRate) = OperatorLib.updateClusterOperators(

then when you go to see if its liquidatable you are not using the eth amounts!!! you are using the old ssv burn rate yikes :)

andrew-blox:
in migrateClusterToETH there can be small enhancement

basically no matter what you call updateClusterOperators

now when it comes to dao you have
        if (!isLiquidated) {
            sp.updateDAOSSV(false, cluster.validatorCount);
        }

you shouldnt need to call updateClusterOperators if isLiquidated = true , just like the dao the operators should already be updated with correct accounting

I know you might say updateClusterOperators returns burnRate and we need that but refer to bug above you dont need the ssv burn rate for any reasons moving forward.

andrew-blox:
i see you have 2 storages for clusterData (ssv and eth)
bytes32 clusterData = s.ethClusters[hashedCluster];
bytes32 clusterDataSSV = s.clusters[hashedCluster];

I thought that eth cluster data just overrides ssv data, this way brings unnecessary gas costs IMO. Maybe i am not seeing full picture. (i put this comment before but deleted it by accident)

mtabasco:
> in validateClusterOnRegistration you have
> 
> ```
>     // todo owner can override ssv cluster here, refactor this check
>     if (clusterData == bytes32(0) && clusterDataSSV!= bytes32(0)) {
>         revert ISSVNetworkCore.IncorrectClusterVersion();
>     }
> ```
> 
> I can technically migrate without calling migrate function. I can liquidate myself / remove validators (clusterData == bytes32(0)) then register a validator and i will migrate. BUt since i didnt officially migrate im not sure we will pay the operators correctly i would assume we pay the operators ssv but with amounts of .00x as in the accounting would pay eth balances but in ssv.
> 
> This needs to be reviewed to make sure its true, tight unit testing here would be great

@andrew-blox This flow can not be performed because of these reasons:
- Validator removal is not possible for SSV clusters. No matter the order of the action (before/after liquidation).
- Even with all cluster values set to 0 and `active = false`, the hashed version is not equal to `bytes32(0)`. So `keccak256(abi.encodePacked(0,0,0,0,false)) != bytes32(0)`

mtabasco:
> i see we have isLiquidatable and isLiquidatableWithEB when registering a validator we call isLiquidatable should we instead be calling isLiquidatableWithEB? we call isLiquidatable as of now in the register validator funct (updateClusterOnRegistration)
> 
> Also isLiquidatable get call should prob call isLiquidatableWithEB internally I didnt check but also reactivate and withdraw should be affected and migrate

Amazing catch @andrew-blox 
I changed `isLiquidatable` by `isLiquidatableWithEB` in `isLiquidatableWithEB`. Thanks!

mtabasco:
> in updateClusterOperators we have if (increaseValidatorCount && !isClusterLiquidated) { operator.validatorCount -= deltaValidatorCount; }
> 
> shouldnt it be += as increaseValidatorCount = true??

The `-=` is correct because during migration, validators are leaving the SSV side (decrease `operator.validatorCount`) and increasing the ETH side (`operator.ethValidatorCount`).

This code block only executes during `migrateClusterToETH` because:

`liquidate`, `reactivate` and `(bulk)removeValidator` require `VERSION_ETH` → operators already have ETH snapshots

mtabasco:
> in _updateOperatorVUnits you have if (deltaPositive) seb.operatorEthVUnits[operatorId] += deltaAbs; else seb.operatorEthVUnits[operatorId] -= deltaAbs
> 
> im not 100% understanding but just want to make sure deltaAbs can never be higher than seb.operatorEthVUnits[operatorId] right?

Correct:
- Each cluster's vUnits contribution to operatorEthVUnits is tracked.
- When decreasing, we only subtract the delta for that specific cluster.
- The operator's total should always be >= any single cluster's contribution.

Please let me know if this answers your question or if I misunderstood. thanks!

mtabasco:
> I think maybe theres some very edge case race condition, in updateBalanceWithEB you have uint64 vUnits = getVUnits(clusterId, cluster.validatorCount);
> 
> Now if the oracle calls to update some cluster by accident calls twice same update func on same cluster really fast in same block, i dont know if the getVunits will give old or updated vunits and how will it apply to the balance.
> 
> Maybe theres nothing wrong here i think you guys should verify though

Good catch. We have active prevention mechanisms for this:
- _verifyEBUpdateFrequency
- _verifyEBStaleness
- Apart from the previous 2:
First call uses `oldVUnits` (line 513) to calculate fees, then sets `newVUnits` 
Second call reads the updated vUnits as `oldVUnits`, calculates fees based on that, then sets another `newVUnits`

mtabasco:
> migrateClusterToETH has 2 major issues first is already stated before using isLiquidatable instead of the eth one
> 
> Second is you get the burn rate at the beginning (uint64 clusterIndex, uint64 burnRate) = OperatorLib.updateClusterOperators(
> 
> then when you go to see if its liquidatable you are not using the eth amounts!!! you are using the old ssv burn rate yikes :)

About the second:
the `burnRate` returned is the ETH fee, not the SSV fee. It uses `operator.ethFee` which is either:

- the operator's declared ETH fee
- the default one (using `defaultOperatorEthFee()`)

About the first issue, I need to confirm something and go back to it.

mtabasco:
> i see you have 2 storages for clusterData (ssv and eth) bytes32 clusterData = s.ethClusters[hashedCluster]; bytes32 clusterDataSSV = s.clusters[hashedCluster];
> 
> I thought that eth cluster data just overrides ssv data, this way brings unnecessary gas costs IMO. Maybe i am not seeing full picture. (i put this comment before but deleted it by accident)

With one single `clusters` mapping, it's hard to identify the type of the cluster (SSV/ETH) without adding extra data (bit flag in the hash, extra mapping to identify the type...), and also, the data inside an SSV/ETH cluster refers to different index types, but from an external point of view, we can not distinguish.
At the end, when all clusters were migrated to ETH, we will use one mapping only and can remove the version checking logic easily, making it efficient again.

andrew-blox:
inside migrateClusterToETH you do

s.ethClusters[hashedCluster] = cluster.hashClusterData();

You never update the ssv hashedCluster (as far as i can see).  Im not saying there is a bug because you dont do this but should it be updated?  Can it lead to a bug later on?  I just imagined that we would hardcode the ssv data to 0 or something thats updated or even better delete the entry to save get back some gas :man_shrugging:

andrew-blox:
in _bulkRemoveValidator you have the below, and single remove revertIfValidatorMissing is true and bulk its false

            if (revertIfValidatorMissing && validatorData == bytes32(0)) {
                revert ISSVNetworkCore.ValidatorDoesNotExist();
            }


why do we have this?  

IMagine i register 10 validators then i go to bulk remove 8 but 4 of them i put in are fake then doesnt the accounting adjust for 8 removals instead of 4?  I get the oracle will update it evuentally but its just weird to allow this, i guess im missing something.

I would imagine if we remove something we dont own then we dont allow

andrew-blox:
small one i guess you just havent updated yet in liquidateSSV you have isLiquidatableWithEB it should instead be isLiquidatable or whatever the ssv function is called i guess 🤷‍♂️

mtabasco:
> inside migrateClusterToETH you do
> 
> s.ethClusters[hashedCluster] = cluster.hashClusterData();
> 
> You never update the ssv hashedCluster (as far as i can see). Im not saying there is a bug because you dont do this but should it be updated? Can it lead to a bug later on? I just imagined that we would hardcode the ssv data to 0 or something thats updated or even better delete the entry to save get back some gas 🤷‍♂️

Sure, we added the `delete` for `s.clusters[hashedCluster]`

mtabasco:
> in _bulkRemoveValidator you have the below, and single remove revertIfValidatorMissing is true and bulk its false
> 
> ```
>         if (revertIfValidatorMissing && validatorData == bytes32(0)) {
>             revert ISSVNetworkCore.ValidatorDoesNotExist();
>         }
> ```
> 
> why do we have this?
> 
> IMagine i register 10 validators then i go to bulk remove 8 but 4 of them i put in are fake then doesnt the accounting adjust for 8 removals instead of 4? I get the oracle will update it evuentally but its just weird to allow this, i guess im missing something.
> 
> I would imagine if we remove something we dont own then we dont allow

Right. We removed that extra check. We have one next that `if` that verifies the state of the validator.

mtabasco:
> small one i guess you just havent updated yet in liquidateSSV you have isLiquidatableWithEB it should instead be isLiquidatable or whatever the ssv function is called i guess 🤷‍♂️

good catch! done.

PR #358, opened 2026-01-11
fix: prevent effective balance precision loss in roundtrip conversion

olegshmuelov:
## Summary

Fix precision loss when converting between `effectiveBalance` and `vUnits`.

**Problem:** `515 ETH → 160937 vUnits → 514 ETH` (lost 1 ETH due to floor division)

**Fix:** Use ceiling division on write path via `Math.ceilDiv`

**Result:** `515 ETH → 160938 vUnits → 515 ETH` ✓

## Changes

- Add `ebToVUnits()` and `vUnitsToEB()` functions to `SSVStorageEB.sol`
- Use these functions in `SSVClusters.sol` and `SSVViews.sol`
- Add unit tests for roundtrip conversion

### DIFF
--- a/contracts/libraries/SSVStorageEB.sol
+++ b/contracts/libraries/SSVStorageEB.sol
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 pragma solidity 0.8.24;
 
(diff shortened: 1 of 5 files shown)


mtabasco:
updated to https://github.com/ssvlabs/ssv-network/pull/360

PR #360, opened 2026-01-13
Fix/effective balance precision loss

mtabasco:
Coming from https://github.com/ssvlabs/ssv-network/pull/358


contracts/libraries/ClusterLib.sol
```diff
@@ -229,4 +229,21 @@ library ClusterLib {
...
+
+    /// @notice Convert effective balance to vUnits using ceiling division (write path)
+    /// @param effectiveBalance The effective balance in ETH
+    /// @return vUnits value with VUNITS_PRECISION scaling
+    function ebToVUnits(uint32 effectiveBalance) internal pure returns (uint64) {
+        uint256 vUnits = uint256(effectiveBalance) * VUNITS_PRECISION;
+        uint256 vUnitsPerValidator = DEFAULT_EB_PER_VALIDATOR / 1 ether;
+        
+        return uint64(vUnits == 0 ? 0 : (vUnits - 1) / vUnitsPerValidator + 1);
+    }
```
olegshmuelov, line 241: I noticed the OpenZeppelin `Math.ceilDiv` approach from the original PR #358 might have been lost during merge conflict.
Consider using it - it's already a project dependency and more readable:

```solidity
import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";

/// @notice Convert effective balance to vUnits using ceiling division (write path)
/// @param effectiveBalance The effective balance in ETH
/// @return vUnits value with VUNITS_PRECISION scaling
function ebToVUnits(uint32 effectiveBalance) pure returns (uint64) {
    return uint64(Math.ceilDiv(
        uint256(effectiveBalance) * VUNITS_PRECISION,
        DEFAULT_EB_PER_VALIDATOR / 1 ether
    ));
}
```
Benefits:
- Self-documenting (clearly ceiling division)
- Audited, battle-tested
- Handles zero case automatically
- OpenZeppelin is already used in the project

contracts/libraries/ClusterLib.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity 0.8.24;
 import "../interfaces/ISSVNetworkCore.sol";
 import {StorageData} from "./SSVStorage.sol";
 import {StorageProtocol} from "./SSVStorageProtocol.sol";
-import {SSVStorageEB, StorageEB, VUNITS_PRECISION} from "./SSVStorageEB.sol";
+import {DEFAULT_EB_PER_VALIDATOR, SSVStorageEB, StorageEB, VUNITS_PRECISION} from "./SSVStorageEB.sol";
```
olegshmuelov, line 7: nit: consider ordering imports as types first, then constants:
```solidity
import {SSVStorageEB, StorageEB, DEFAULT_EB_PER_VALIDATOR, VUNITS_PRECISION} from "./SSVStorageEB.sol";
```

contracts/libraries/ClusterLib.sol
mtabasco: hey @olegshmuelov, yes, this was intended. Although importing the library does not increase the contract size too much, as the formula is simple, we can micro-optimize this part by copy-pasting the contents.

contracts/libraries/ClusterLib.sol
mtabasco: Sure. We will lint all the code soon.

yurii-ssv (approved):

PR #362, opened 2026-01-13
feat: allow multiple unstake requests

yurii-ssv:


contracts/modules/SSVStaking.sol
```diff
@@ -59,37 +63,76 @@ contract SSVStaking is ISSVStaking, SSVReentrancyGuard {
...
+            revert UnstakeAmountExceedsBalance();
+        }
+
+        EnumerableMap.UintToUintMap storage requests = s.withdrawalRequests[msg.sender];
+        bool wasEmpty = requests.length() == 0;
 
-        // todo maybe use blocks here
         uint64 unlockTime = uint64(block.timestamp + s.cooldownDuration);
-        s.withdrawals[msg.sender] = UnstakeRequest({amount: uint192(amount), unlockTime: unlockTime});
+        requests.set(amount, unlockTime);
```
venimir-ssv: using amount as key here might override request if the user unstake with same amount twice
I would suggest to make a struct and use global variable for requestId

`struct WithdrawalRequest { uint256 amount; uint64 unlockTime; }
uint256 requestId = s.nextWithdrawalId++;
bytes req = abi.encode(WithdrawRequest{amount, unlocktime})
requests.set(requestId, req)
`

venimir-ssv (approved):

contracts/modules/SSVStaking.sol
```diff
@@ -59,37 +59,66 @@ contract SSVStaking is ISSVStaking, SSVReentrancyGuard {
...
-        }
-
         _syncFees(s);
 
         uint256 bal = ICSSVToken(cssv).balanceOf(msg.sender);
         _settleWithBalance(msg.sender, bal, s);
-        if (amount > bal) revert UnstakeAmountExceedsBalance();
 
-        _removeDelegation(msg.sender, amount, bal, s);
+        uint256 totalRequested = calculateTotalRequestedBalance(s) + amount;
```
mtabasco: @yurii-ssv As the total requested is already burned (line 80), summing total requested + amount will result in an amount that will likely be greater than the cSSV balance.

Checking `if (amount > bal) ` should be enough.

mtabasco: @yurii-ssv As the total requested is already burned (line 80), summing total requested + amount will result in an amount that will likely be greater than the cSSV balance.

Checking `if (amount > bal) ` should be enough.

contracts/modules/SSVStaking.sol
```diff
@@ -59,37 +59,66 @@ contract SSVStaking is ISSVStaking, SSVReentrancyGuard {
...
+        requests.push(UnstakeRequest({amount: uint192(amount), unlockTime: unlockTime}));
+
+        _removeDelegation(msg.sender, amount, bal, s);
+
+        ICSSVToken(cssv).burn(msg.sender, amount);
 
         emit UnstakeRequested(msg.sender, amount, unlockTime);
     }
 
+    function calculateTotalRequestedBalance(StorageStaking storage s) internal view returns (uint256) {
```
mtabasco: this can be removed

contracts/modules/SSVStaking.sol
```diff
@@ -59,37 +59,66 @@ contract SSVStaking is ISSVStaking, SSVReentrancyGuard {
...
+        uint256 total = 0;
+        UnstakeRequest[] storage requests = s.withdrawalRequests[msg.sender];
+        for (uint256 j = 0; j < requests.length; j++) {
+            total += requests[j].amount;
+        }
+        return total;
+    }
+
+    function calculateTotalUnfrozenBalance(StorageStaking storage s) internal returns (uint256) {
```
mtabasco, line 88:
```solidity
// OPTIMIZATION: Single pass, no memory allocation
function calculateTotalUnfrozenBalance(StorageStaking storage s) internal returns (uint256) {
    UnstakeRequest[] storage requests = s.withdrawalRequests[msg.sender];
    uint256 total = 0;
    uint256 i = 0;
        
    // Swap-and-pop pattern
    while (i < requests.length) {
        if (requests[i].unlockTime <= block.timestamp) {
            total += requests[i].amount;
            // Move last element to current spot and pop last
            requests[i] = requests[requests.length - 1];
            requests.pop();
            // Do not increment i, checking the swapped element next
        } else {
            i++;
        }
    }
    return total;
}
```

contracts/modules/SSVStaking.sol
```diff
@@ -59,37 +59,66 @@ contract SSVStaking is ISSVStaking, SSVReentrancyGuard {
...
 
-        if (s.withdrawals[msg.sender].amount != 0) {
-            revert CooldownActive();
-        }
-
         _syncFees(s);
 
         uint256 bal = ICSSVToken(cssv).balanceOf(msg.sender);
         _settleWithBalance(msg.sender, bal, s);
-        if (amount > bal) revert UnstakeAmountExceedsBalance();
```
mtabasco, line 67: Although in reality users will not have more than 3 or 4 requests in parallel, I think it's good to have a limit like 10 MAX_PENDING_REQUESTS.
`if (requests.length >= MAX_PENDING_REQUESTS) revert;`

mtabasco (changes requested):
@yurii-ssv good job! I pointed to a bug and add small optimization

mtabasco (approved):
GO!

PR #363, opened 2026-01-14
gas profile and gas stats commands added

venimir-ssv:


test/helpers/gas-usage.ts
```diff
@@ -0,0 +1,294 @@
...
+  REQUEST_UNSTAKE,
+  WITHDRAW_UNSTAKE,
+  CLAIM_ETH_REWARDS,
+  SYNC_FEES,
+  RESCUE_ERC20,
+}
+
+const MAX_GAS_PER_GROUP: any = {
+  /* REAL GAS LIMITS - adjusted for harness and integration contracts */
+  [GasGroup.REGISTER_OPERATOR]: 210000,
```
mtabasco, line 120: @venimir-ssv I see something strange here:
On V1 we have:
`[GasGroup.REGISTER_OPERATOR]: 137000`
 and here:
`[GasGroup.REGISTER_OPERATOR]: 210000`
Which I assume is normal because we don't use a high optimizer value and use new data structures, etc

But here on V1 and V2 it's the same value:
`[GasGroup.REGISTER_VALIDATOR_EXISTING_CLUSTER]: 202000`
So even by changing the optimizer value, this one should be higher on V2.

Did all the previous gas trackers execute?

test/helpers/gas-usage.ts
```diff
@@ -0,0 +1,294 @@
...
+  [GasGroup.UPDATE_CLUSTER_BALANCE]: 300000,
+
+  [GasGroup.SET_UNSTAKE_COOLDOWN]: 80000,
+  [GasGroup.SET_QUORUM]: 80000,
+  [GasGroup.REPLACE_ORACLE]: 120000,
+  [GasGroup.COMMIT_ROOT]: 150000,
+  [GasGroup.NETWORK_FEE_CHANGE_SSV]: 50000,
+  [GasGroup.WITHDRAW_NETWORK_SSV_EARNINGS]: 80000,
+
+  [GasGroup.STAKE_SSV]: 400000,
```
mtabasco, line 221: woah 400k for a stake operation...we need to optimize, no prob.
@venimir-ssv can you please add another gas group to measure a 2nd stake operation for the same user? like `REGISTER_VALIDATOR_EXISTING_CLUSTER`, but for staking, like INITIAL_STAKE_SSV, POST_INITIAL_STAKE_SSV

mtabasco (changes requested):
@venimir-ssv great!
I leave some comments. thanks!

test/helpers/gas-usage.ts
venimir-ssv: POST_INITIAL_STAKE_SSV added and the metric is 134K

venimir-ssv: https://github.com/ssvlabs/ssv-network/pull/363/commits/9f44440de5978d02aec6d6ee93a5fdbb651705f6

mtabasco (approved):

PR #366, opened 2026-01-16
Fix/removed operator

venimir-ssv:


andrew-blox:
If you call removeOperator then you make operator.block = 0.  Now if i call removeValidator and that validator is in a cluster of that removed operator, eventually the function "updateSnapshotSt" or "updateSnapshotSSV" is called and either way both of these functions update operator.block with the current block with no checks to see if its removed.

Now what are the repercussions to this?  Im not sure lol but im assuming you made the block 0 on removal for a reason that it cant be updated.

andrew-blox:
in _executeLiquidation you have 
        if (vUnitsCluster > 0) {
            uint64 baselineVUnits = uint64(cluster.validatorCount) * VUNITS_PRECISION;

            if (vUnitsCluster != baselineVUnits) {
                bool moreThanBaseline = vUnitsCluster > baselineVUnits;

Sorry maybe im mistaken but if moreThanBaseline == false then we have a very big issue no?  How can baselineVUnits ever be more than vUnitsCluster, remember you check for equality on the line above it so it really is baseline more than vunits which is impossible right?

Ok so i know how you can if its ssv cluster, i really gotta say the way this is all being done between the versions is very odd to me, maybe we can have quick chat on what i see as a bit more efficient here.

andrew-blox:
inside _executeLiquidation you have 
                    if (version == CoreLib.VERSION_ETH) {
                        if (moreThanBaseline) sp.daoTotalEthVUnits -= delta;
                        else sp.daoTotalEthVUnits += delta;
                    } else {
                        if (moreThanBaseline) sp.daoTotalVUnits -= delta;
                        else sp.daoTotalVUnits += delta;
                    }

basically if its ssv cluster you do + or - to sp.daoTotalVUnits, theres vunits in dao ssv?  what is this referring to? doesnt ssv dao just work with validator count?

same thing here below, what is operatorVUnits in ssv?

if (version == CoreLib.VERSION_ETH) seb.operatorEthVUnits[opId] -= vUnitsCluster;
else seb.operatorVUnits[opId] -= vUnitsCluster;

I guess the question also is when we liquidate eth cluster then reactivate we use your last eff balances not 32 * val count so why dont we do the same with liquidate ssv then user does migrate we should calculate based on last updated eff balance not 32*val count right?  Im not sure im just asking <- update, i talked with Ariel and he said it should not be 32 * val count it should not clear eff balance and when you do migrate you should use the eff balance

andrew-blox:
in liquidateSSV you call isLiquidatableWithEB shouldnt you call isLiquidatable?  I think both will actually work just wondering why keep isLiquidatable, maybe im missing something?

PR #374, opened 2026-01-27
Fix/operator fee change

mtabasco:


contracts/modules/SSVOperators.sol
```diff
@@ -13,12 +13,17 @@ import {Counters} from "@openzeppelin/contracts/utils/Counters.sol";
...
     uint64 private constant PRECISION_FACTOR = 10_000;
+    uint256 public immutable UPGRADE_TIMESTAMP;
 
     using Types256 for uint256;
     using Types64 for uint64;
     using Counters for Counters.Counter;
     using OperatorLib for Operator;
 
+    constructor(uint256 upgradeTimestamp) {
+        UPGRADE_TIMESTAMP = upgradeTimestamp;
```
mtabasco, line 24: @venimir-ssv what do you think of using using `block.timestamp` ?
I put a parameter here to allow the DAO to put a specific line

venimir-ssv: I would go with block.number but timestamp is ok as well

yurii-ssv (approved):

PR #386, opened 2026-02-03
Feature/deprecate delegations

yurii-ssv:


contracts/interfaces/ISSVNetworkCore.sol
```diff
@@ -79,7 +79,7 @@ interface ISSVNetworkCore {
     error InvalidOperatorIdsLength(); // 0x38186224
     error ClusterAlreadyEnabled(); // 0x3babafd2
     error ClusterIsLiquidated(); // 0x95a0cf33
-    error ClusterDoesNotExists(); // 0x185e2b16
+    error ClusterDoesNotExist(); // 0x185e2b16
```
venimir-ssv, line 82: we need to update the hash sig here as well

yurii-ssv: Agree, but I think it should be done with natspec update

venimir-ssv (approved):

contracts/modules/SSVStaking.sol
mtabasco: @yurii-ssv remove unused import

contracts/modules/SSVViews.sol
mtabasco: @yurii-ssv remove unused import

mtabasco (approved):

PR #398, opened 2026-02-11
fix: use EB-weighted balance settlement in registration and removal

liorrutenberg:
Registration (updateClusterOnRegistration) and removal (_bulkRemoveValidator) were settling accrued fees using flat updateBalance() which multiplies by validatorCount, ignoring effective balance. Clusters with high EB were systematically under-charged. SSVViews.isLiquidatable() also settled balance with the non-EB method.

- Make updateClusterData() call updateBalanceWithEB() instead of updateBalance()
- Pass hashedCluster through registration, removal and view call sites
- Remove dead updateBalance() function (no remaining callers)
- Update echidna test harness to match new signatures
- Add unit tests verifying EB-weighted settlement on register and remove


mtabasco:
Replaced by https://github.com/ssvlabs/ssv-network/pull/400

PR #399, opened 2026-02-11
fix: migrated operator accounting

yurii-ssv:


venimir-ssv (approved):

venimir-ssv (approved):

contracts/libraries/OperatorLib.sol
```diff
@@ -360,23 +360,25 @@ library OperatorLib {
             uint64 operatorId = operatorIds[i];
             ISSVNetworkCore.Operator storage operator = s.operators[operatorId];
 
+            // Update SSV snapshot before validator count changes
```
mtabasco, line 363: @yurii-ssv this change prevents using wrong validatorCount when updating SSV snapshot

PR #410, opened 2026-02-11
fix: forbid to create clusters with a removed operator

yurii-ssv:


venimir-ssv (approved):

contracts/libraries/OperatorLib.sol
```diff
@@ -183,6 +185,20 @@ library OperatorLib {
             ensureETHDefaults(s.operators[operatorId]);
```
olegshmuelov: **Bug: existing-cluster path doesn't catch removed operators.**

`_resetOperatorState()` does NOT reset `owner`, so for a removed operator `owner != address(0)` is always true. The `else` branch (existing cluster, line 197) only checks `owner == address(0)`, meaning a removed operator passes the check and gets revived with `ethFee=0`.

**Repro:** Register cluster `[1,2,3,4]` → remove operator 3 → add more validators to the same cluster → operator 3 is silently revived.

**Simpler fix** — replace the entire `if/else` block with a single check BEFORE `ensureETHDefaults` (no `isExistingCluster` param needed):

```solidity
if (s.operators[operatorId].snapshot.block == 0 && 
    s.operators[operatorId].ethSnapshot.block == 0) {
    revert ISSVNetworkCore.OperatorDoesNotExist();
}
ensureETHDefaults(s.operators[operatorId]);
```

This catches removed operators (both blocks == 0) for both new and existing clusters. Also eliminates the `isExistingCluster` parameter from `OperatorLib`, `ClusterLib`, and `SSVValidators` — 3-line fix with zero interface changes.

test/integration/SSVNetwork.test.ts
```diff
@@ -1734,6 +1734,26 @@ describe("SSVNetwork full integration tests", () => {
...
 
+    it("Is reverted with 'OperatorDoesNotExist' if one of operators is removed", async function () {
+      const { network } =
+        await networkHelpers.loadFixture(deployFullSSVNetworkFixture);
+
+      const validatorKey = makePublicKey(1);
+      const operatorIds = await registerOperators(network, operatorOwner, 4);
+      await whitelistAddresses(network, operatorOwner, operatorIds, [clusterOwner.address]);
+
+      await network.connect(operatorOwner).removeOperator(operatorIds[2]);
```
olegshmuelov, line 1745: **Missing test case:** This only tests new cluster registration (`EMPTY_CLUSTER`). Need a second test that registers validators to an **existing** cluster after an operator is removed — that's the path the fix doesn't cover. Something like:

```typescript
// Register first validator (creates cluster)
const cluster = await registerValidator(...);
// Remove operator
await network.connect(operatorOwner).removeOperator(operatorIds[2]);
// Add second validator to EXISTING cluster — should also revert
await expect(network.connect(clusterOwner).registerValidator(
  validatorKey2, operatorIds, DEFAULT_SHARES, cluster,
  { value: DEFAULT_ETH_REGISTER_VALUE }
)).to.be.revertedWithCustomError(network, Errors.OPERATOR_DOES_NOT_EXIST);
```

yurii-ssv: Registering validators to a cluster with a removed operator is allowed, this is by design

yurii-ssv: for existing clusters**

mtabasco (approved):

olegshmuelov:
Guard before `ensureETHDefaults` and `&&` logic are correct for new clusters. But existing-cluster `else` branch still only checks `owner == address(0)` — removed operators pass because `_resetOperatorState` doesn't reset `owner`.

Register cluster → remove op 3 → add validator to same cluster → op 3 revived with `ethFee=0`, `ethValidatorCount > 0`. Operator didn't consent to revival and earns nothing. If operator owner re-removes, it re-creates the scenario that #412 fixes.

Same `(ethSnapshot.block == 0 && snapshot.block == 0)` check should apply to both paths — the `isExistingCluster` split isn't needed.

andrew-blox:
Ok now i was not in this too deep into this bug and just read over it, maybe im wrong but to me this all comes down to 1 line of code that can fix this, its also something i have opened a bug on twice but also was told many times that it doesnt affect anything but i really agree it has unknowns and is just easily solvable, this line is this

inside updateSnapshotSt func you have the below
operator.ethSnapshot.block = currentBlock;

this is what will recreate the operator.  As said before we should have simple if to check that the operator is not removed before doing this (reviving the operator).

We have not done any testing on this so called revived operator to see how giving it a block will affect it.

Now as for accounting i know that this removed operator wont get balances because his fee is 0 but im not fully aware of all the implications or edgey things you can do by giving a removed operator a revived block.

@olegshmuelov maybe im wrong but would the fix above fix your issue?  Like when you migrate if an operator(s) is removed we will give it a block but because his ssv fee is 0 his eth fee will also be 0.

olegshmuelov:
> Ok now i was not in this too deep into this bug and just read over it, maybe im wrong but to me this all comes down to 1 line of code that can fix this, its also something i have opened a bug on twice but also was told many times that it doesnt affect anything but i really agree it has unknowns and is just easily solvable, this line is this
> 
> inside updateSnapshotSt func you have the below operator.ethSnapshot.block = currentBlock;
> 
> this is what will recreate the operator. As said before we should have simple if to check that the operator is not removed before doing this (reviving the operator).
> 
> We have not done any testing on this so called revived operator to see how giving it a block will affect it.
> 
> Now as for accounting i know that this removed operator wont get balances because his fee is 0 but im not fully aware of all the implications or edgey things you can do by giving a removed operator a revived block.
> 
> @olegshmuelov maybe im wrong but would the fix above fix your issue? Like when you migrate if an operator(s) is removed we will give it a block but because his ssv fee is 0 his eth fee will also be 0.

You're right that `operator.ethSnapshot.block = currentBlock` in updateSnapshot is what revives the operator. But fixing it inside updateSnapshot is tricky — you can't distinguish a removed operator (block==0) from a first-time ETH operator (block==0 but legitimate). Both have ethSnapshot.block == 0 at that point.

The cleanest fix is adding the same snapshot check to the existing-cluster else branch in updateClusterOperatorsOnRegistration (OperatorLib.sol:195-199):

```solidity
} else {
    if (operator.owner == address(0) ||
        (operator.ethSnapshot.block == 0 && operator.snapshot.block == 0)) {
        revert ISSVNetworkCore.OperatorDoesNotExist();
    }
}
```

This is identical to the new-cluster check above it. Catches the removed operator before ensureETHDefaults and updateSnapshot ever run.

Migration path is already handled separately (PR #401 skips removed operators at line 380).

andrew-blox:
Ok i think i finally found a big issue in this talk, i think weve just been looking for implications to this and heres an example

1. create operator A
2. register validators to this operator
3. remove operator
4. add validators (reviving ssv block on the operator)
5. migrate contract 
6. add validators (reviving eth block on the operator)
7. its now impossible to liquidate this cluster due to underflow on this line

            for (uint256 i; i < operatorIds.length; ++i) {
                ISSVOperators.Operator storage op = s.operators[operatorIds[i]];
                if (op.ethSnapshot.block != 0 && op.snapshot.block != 0) { <- THIS PASSES
                    op.ethValidatorCount -= cluster.validatorCount; <- UNDERFLOW
                }
            }

olegshmuelov:
> Ok i think i finally found a big issue in this talk, i think weve just been looking for implications to this and heres an example
> 
> 1. create operator A
> 2. register validators to this operator
> 3. remove operator
> 4. add validators (reviving ssv block on the operator)
> 5. migrate contract
> 6. add validators (reviving eth block on the operator)
> 7. its now impossible to liquidate this cluster due to underflow on this line
>    ```
>         for (uint256 i; i < operatorIds.length; ++i) {
>             ISSVOperators.Operator storage op = s.operators[operatorIds[i]];
>             if (op.ethSnapshot.block != 0 && op.snapshot.block != 0) { <- THIS PASSES
>                 op.ethValidatorCount -= cluster.validatorCount; <- UNDERFLOW
>             }
>         }
>    ```

I tested your exact scenario with a PoC. The underflow doesn't happen with the current code. Here's why:

1. You can't register to an SSV-only cluster — reverts with `IncorrectClusterVersion` (ClusterLib.sol:204). So step 4 (before migration) is blocked.

2. If you migrate first, then remove, then register — migration skips removed operators (OperatorLib.sol:380: `if (operator.snapshot.block == 0 && operator.ethSnapshot.block == 0) continue`). So `snapshot.block` stays 0.

3. Registration to existing ETH cluster revives `ethSnapshot.block` but NOT `snapshot.block`. PR #412's guard (`op.ethSnapshot.block \!= 0 && op.snapshot.block \!= 0`) catches this — `snapshot.block` is still 0, so the operator is skipped. No underflow.

I tested all three scenarios — PoC confirms no underflow path exists with current code.

The 32 issue revival itself still happens (removed operator gets ethValidatorCount > 0, ethFee = 0), but the underflow consequence is mitigated by PR #412.

PR #427, opened 2026-02-17
Claude init

liorrutenberg:


docs/SPEC.md
```diff
@@ -0,0 +1,836 @@
...
+
+- Operator fees paid in ETH
+- Network fees paid in ETH
+- ETH deposited upfront for runway
+- Fees scale with effective balance (vUnits), not validator count
+
+### Existing Clusters (SSV-based, Legacy)
+
+- Continue running with existing SSV runway
+- **Blocked operations**: add validators, remove validators, reactivate, deposit SSV
```
arielssv: @liorrutenberg + withdrawing SSV

docs/SPEC.md
```diff
@@ -0,0 +1,836 @@
...
+- **Allowed operations**: self-liquidate, migrate to ETH, exit validators
+- SSV fee accrual continues normally until runway depletes or migration occurs
+
+### Cluster Migration (`migrateClusterToETH`)
```
arielssv, line 218: @liorrutenberg reactivates an liquidated cluster & emits reactivate event in addition to migration

docs/SPEC.md
```diff
@@ -0,0 +1,836 @@
...
+| `reactivate(..., uint256 amount, Cluster)` | `reactivate(..., Cluster) payable` | `amount` removed, now `payable` |
+| `getBalance(...) returns (uint256)` | `getBalance(...) returns (uint256, uint256)` | Now also returns `ebBalance` |
+
+---
+
+## 2. Effective Balance Accounting
+
+### Overview
+
+Fees are calculated based on a cluster's total effective balance (in whole ETH) rather than validator count. This supports post-Pectra validators with variable effective balances (32–2048 ETH per validator).
```
arielssv: @liorrutenberg  not sure about the "(in whole ETH)" part

docs/FLOWS.md
```diff
@@ -0,0 +1,977 @@
...
+   - Increment `operator.ethValidatorCount`
+   - If first ETH interaction: `ensureETHDefaults()` sets ethFee and ethSnapshot.block
+2. Store validator: `validatorPKs[hash(pubkey, owner)] = hash(operatorIds | active=true)`
+3. Update cluster state:
+   - `cluster.validatorCount++`
+   - `cluster.balance += msg.value`
+   - `cluster.index = current cumulative operator ETH index`
+   - `cluster.networkFeeIndex = current ETH network fee index`
+4. Update DAO: `ethDaoValidatorCount++`, `daoTotalEthVUnits += VUNITS_PRECISION` (baseline 32 ETH)
+5. If cluster has explicit EB: update operator and DAO deviation vUnits
```
arielssv: @liorrutenberg these are ETH clusters, why dont we update cluster & operators & DAO EB regardless by 32 * validator count?

docs/SPEC.md
```diff
@@ -0,0 +1,836 @@
...
+
+---
+
+## 1. ETH Payments
+
+### Overview
+
+ETH replaces SSV as the payment asset for network and operator fees. All new clusters operate exclusively with ETH. Existing SSV clusters are legacy — they cannot add/remove validators, deposit SSV, or reactivate. The only forward path is migration to ETH.
+
+### New Clusters (ETH-based)
```
arielssv, line 203: add operates with EB accounting

docs/SPEC.md
```diff
@@ -0,0 +1,836 @@
...
+
+- Operator fees paid in ETH
+- Network fees paid in ETH
+- ETH deposited upfront for runway
+- Fees scale with effective balance (vUnits), not validator count
+
+### Existing Clusters (SSV-based, Legacy)
```
arielssv, line 211: add operates with linear validator count accounting

docs/SPEC.md
```diff
@@ -0,0 +1,836 @@
...
+|---|---|---|
+| `registerValidator(..., uint256 amount, Cluster)` | `registerValidator(..., Cluster) payable` | `amount` removed, now `payable` |
+| `bulkRegisterValidator(..., uint256 amount, Cluster)` | `bulkRegisterValidator(..., Cluster) payable` | `amount` removed, now `payable` |
+| `deposit(..., uint256 amount, Cluster)` | `deposit(..., Cluster) payable` | `amount` removed, now `payable` |
+| `reactivate(..., uint256 amount, Cluster)` | `reactivate(..., Cluster) payable` | `amount` removed, now `payable` |
+| `getBalance(...) returns (uint256)` | `getBalance(...) returns (uint256, uint256)` | Now also returns `ebBalance` |
+
+---
+
+## 2. Effective Balance Accounting
```
arielssv, line 279: it doesnt say anywhere that this is for ETH clusters.

docs/SPEC.md
```diff
@@ -0,0 +1,836 @@
...
+
+VUNITS_PRECISION = 10,000
+```
+
+Examples:
+- 1 validator at 32 ETH → 10,000 vUnits
+- 1 validator at 64 ETH → 20,000 vUnits
+- 3 validators at 32 ETH each → 30,000 vUnits
+
+### Implicit vs Explicit EB
```
arielssv, line 301: this is a bit misleading because although we update EB for legacy clusters, they still use the previous accounting.

docs/SPEC.md
```diff
@@ -0,0 +1,836 @@
...
+- Mint: only by SSVStaking on `stake()`
+- Burn: only by SSVStaking on `requestUnstake()`
+- Transfer hook: `_beforeTokenTransfer` calls `SSVStaking.onCSSVTransfer(from, to, amount)`
+  - Settles rewards for both sender and receiver before transfer
+  - Ensures rewards accrued up to transfer point stay with original holder
+- Retains full DAO governance voting power
+
+### Unstaking (Two-Step)
+
+1. **`requestUnstake(amount)`**: Burns cSSV, creates `UnstakeRequest{amount, unlockTime = now + cooldownDuration}`. Max 10 pending requests per user.
```
arielssv: Stakers may submit multiple withdrawal requests over time. When finalizing an unstake, the staker can claim the cumulative amount of all requests whose lock period has fully elapsed, while any requests still in their lock period remain locked. A maximum of 2,000 active withdrawal requests per staker is supported.

docs/SPEC.md
```diff
@@ -0,0 +1,836 @@
...
+
+---
+
+## 4. Oracle System
+
+### Overview
+
+Effective Balance Oracles track validator balances on the beacon chain and commit Merkle roots on-chain. The protocol uses a permissioned set of 4 oracles with a 3-of-4 (75%) quorum threshold.
+
+### Commit Flow (`commitRoot`)
```
arielssv, line 408: missing reference for the last commit (outside qurom)

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+Same as 1.1 but for multiple validators in one transaction. Each validator emits a separate `ValidatorAdded` event. `msg.value` is added to cluster balance once (not per validator).
+
+#### Additional Invariants
+- `contract.balance == previous + msg.value` (single ETH deposit)
+- `operator.ethValidatorCount == previous + N` for each operator (N = number of validators)
+- `ethDaoValidatorCount == previous + N`
+
+---
+
+### 1.3 Remove Validator
```
arielssv, line 121: missing:
- bulk remove validators
- exit validators
- bulk exit validators

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+---
+
+### 1.5 Withdraw ETH
+
+**Caller:** Cluster owner
+**nonReentrant:** Yes
+
+#### Preconditions
+- Cluster must exist as ETH cluster (VERSION_ETH)
+- Cluster must be active
```
arielssv: false

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+```
+
+---
+
+### 1.5 Withdraw ETH
+
+**Caller:** Cluster owner
+**nonReentrant:** Yes
+
+#### Preconditions
```
arielssv, line 255: contract must have sufficient ETH balance

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+
+**Caller:** Anyone (self-liquidation always allowed; third-party only if cluster is liquidatable)
+**nonReentrant:** Yes
+
+#### Preconditions
+- Cluster must exist as ETH cluster (VERSION_ETH)
+- Cluster must be active
+- If caller != owner: cluster must be liquidatable (balance below threshold)
+
+#### State Mutations
```
arielssv, line 295: reduce operators and dao EB

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+### 1.8 Reactivate
+
+**Caller:** Cluster owner
+**Payable:** Yes (msg.value = ETH deposit)
+
+#### Preconditions
+- Cluster must exist as ETH cluster
+- Cluster must be liquidated (`active == false`)
+
+#### State Mutations
```
arielssv, line 338: increase operators and dao EB

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+
+### 4.2 Remove Operator
+
+**Caller:** Operator owner
+**nonReentrant:** Yes
+
+#### Preconditions
+- Operator must exist (`snapshot.block != 0 || ethSnapshot.block != 0`)
+- Caller must be operator owner
+- Operator must have 0 validators in BOTH SSV and ETH counts
```
arielssv: false

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+1. Store `OperatorFeeChangeRequest{fee: packed(newFee), approvalBeginTime: now + declarePeriod, approvalEndTime: now + declarePeriod + executePeriod}`
+
+#### Events
+```solidity
+emit OperatorFeeDeclared(owner, operatorId, block.number, fee);
+```
+
+---
+
+### 4.4 Execute Operator Fee
```
arielssv, line 633: we need reference to what happens to declare operator fee and execute operator fee a flows between contract upgrades.

what happens to "OperatorFeeChangeRequest"?

there is no validation on fee limits at execute.

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+
+#### Postcondition Invariants
+- `operators[id].owner` preserves the original owner address (non-zero)
+- All other operator fields are zeroed: snapshots, fees, validator counts
+- No earnings remain in the system for this operator
+- Public key can be re-registered
+
+---
+
+### 4.3 Declare Operator Fee
```
arielssv, line 609: missing reference to override when called multiple within period

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+1. Update operator ETH snapshot
+2. Set `operator.ethFee = request.fee`
+3. Delete fee change request
+
+#### Events
+```solidity
+emit OperatorFeeExecuted(owner, operatorId, block.number, fee);
+```
+
+#### Postcondition Invariants
```
arielssv, line 653: this has no impact on their index calcs?

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+
+### 4.5 Reduce Operator Fee
+
+**Caller:** Operator owner (immediate, no timelock)
+
+#### Preconditions
+- New fee within `[minimumOperatorEthFee, currentFee)`
+- New fee strictly less than current
+
+#### State Mutations
```
arielssv, line 668: this has no impact on their index calcs?

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+emit OperatorFeeDeclarationCancelled(owner, operatorId);
+```
+
+#### Postcondition Invariants
+- No pending fee change request for this operator
+- Operator's current fee is unchanged
+
+---
+
+### 4.7 Withdraw Operator Earnings (ETH)
```
arielssv, line 703: dont we also have withdrawAllVersionOperatorEarnings?

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+---
+
+## 5. Staking Flows
+
+### 5.1 Stake SSV
+
+**Caller:** Anyone with SSV tokens
+**nonReentrant:** Yes
+
+#### Preconditions
```
arielssv, line 777: pretty sure this was not implemented, but as a safety mechanism we need to validate that
cSSV.total supply + amount <= SSV.total supply

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+
+### 5.2 Request Unstake
+
+**Caller:** cSSV holder
+**nonReentrant:** Yes
+
+#### Preconditions
+- `amount > 0`
+- `amount <= cSSV.balanceOf(msg.sender)`
+- Pending unstake requests < MAX_PENDING_REQUESTS (10)
```
arielssv: 2000

arielssv (commented):
@liorrutenberg

arielssv (commented):
@liorrutenberg

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+2. Set `operator.ethFee = request.fee`
+3. Delete fee change request
+
+#### Events
+```solidity
+emit OperatorFeeExecuted(owner, operatorId, block.number, fee);
+```
+
+#### Postcondition Invariants
+- `operator.ethFee == packed(newFee)`
```
RaekwonIII: incoherent with line 674:

> 2. Set `operator.ethFee = request.fee`

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+
+#### Preconditions
+- Pending fee change request exists
+- `approvalBeginTime > UPGRADE_TIMESTAMP` (reject pre-migration declarations)
+- Current time within `[approvalBeginTime, approvalEndTime]`
+- Fee still within `operatorMaxFee`
+
+#### State Mutations
+1. Update operator ETH snapshot
+2. Set `operator.ethFee = request.fee`
```
RaekwonIII: incoherent with postcondition

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+- `amount <= accumulated ETH earnings`
+
+#### State Mutations
+1. Update ETH snapshot (accumulate latest earnings)
+2. Deduct `amount` from snapshot balance
+3. Transfer `amount` ETH to operator owner
+
+#### Events
+```solidity
+emit OperatorWithdrawn(owner, operatorId, amount);
```
RaekwonIII: As discussed: `OperatorWithdrawnETH` (or equivalent),  ETH-specific event

docs/SPEC.md
```diff
@@ -0,0 +1,836 @@
...
+| Parameter | Current Value | Proposed Value |
+|---|---|---|
+| `minimumLiquidationCollateralSSV` | 1.53 SSV | 0.883 SSV |
+| `minimumBlocksBeforeLiquidationSSV` | 100,380 (~14 days) | 100,380 (~14 days) |
+
+### Staking Parameters
+
+| Parameter | Initial Value | Update Function |
+|---|---|---|
+| `cooldownDuration` | 604,800 seconds (7 days) | `setUnstakeCooldownDuration(uint64)` |
```
olegshmuelov, line 957: **Severity: MEDIUM** — governance misconfiguration risk

The DIP-X proposal explicitly defines cooldownDuration in **blocks**, not seconds:

> **DIP-X proposal, Governance Parameters table (SSV Staking section):**
> "cooldownDuration — Unstake cooldown duration (**in blocks**): the period users must wait between requesting an unstake and being able to withdraw their unlocked SSV."
> "Update function: setUnstakeCooldownDuration(**uint64 blocks**)"
> "Initial Value: **50120** (7 days)"

The spec says **604,800 seconds**. The code confirms seconds-based implementation:
`uint64 unlockTime = uint64(block.timestamp + s.cooldownDuration)` (SSVStaking.sol:88)

Both target ~7 days but the semantics differ — blocks-based cooldown ties to chain progress, seconds-based is wall-clock time. Also the converted values don't match: 604,800s / 12 = 50,400 blocks ≠ 50,120 blocks.

**Concrete risk:** If governance follows DIP-X and sets 50120 (thinking it's blocks), the actual cooldown would be 50,120 seconds = **~13.9 hours**, not 7 days.

If this deviation from the proposal was intentional, the spec should note it. If not, the spec and code should be updated.

docs/SPEC.md
olegshmuelov: **Severity: MEDIUM** — economic model deviation from proposal

The DIP-X proposal describes a **lock-then-burn** two-step unstaking model:

> **DIP-X proposal, "Unstaking" section:**
> "Unstaking is a two-step process:
> First, the staker submits a withdrawal request, which **locks** the specified amount of cSSV and stops reward accrual for that portion. It is proposed that the protocol will launch with a 7-day lock period.
> Once the lock period ends, the staker can finalize the unstake. The **locked cSSV is burned**, and the underlying SSV is returned at a 1:1 ratio relative to the original stake."

The spec and code implement **burn-then-wait** instead:
- Step 1: cSSV is **burned immediately** at `requestUnstake` (SSVStaking.sol:91 calls `ICSSVToken(CSSV_ADDRESS).burn(msg.sender, amount)`)
- Step 2: SSV tokens returned after cooldown (`withdrawUnlocked` — no cSSV interaction)

This is a material difference:
- **Reward distribution**: Burning immediately reduces `cSSV.totalSupply()`, accelerating `accEthPerShare` for remaining stakers. Under the proposal's lock model, supply wouldn't decrease until finalization.
- **Governance**: Voting power is lost at request time vs finalization time.

If the burn-then-wait model was a deliberate design choice, the spec should note the deviation from the proposal and explain why.

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+
+#### State Mutations (ETH Cluster)
+
+1. Convert `effectiveBalance` to `newVUnits = ebToVUnits(effectiveBalance)`
+2. Compute `effectiveOldVUnits`:
+   - If `storedVUnits == 0`: `validatorCount * VUNITS_PRECISION`
+   - Else: `storedVUnits`
+3. If cluster active: settle operator and network fees using OLD vUnits
+4. If `newVUnits != effectiveOldVUnits` AND cluster active:
+   - For each operator: `operatorEthVUnits[opId] += (newVUnits - effectiveOldVUnits) / operatorCount`
```
olegshmuelov: **Severity: HIGH** — incorrect formula, would break accounting if someone implements from this doc

This formula is wrong — there is no division by operatorCount in the code.

SSVClusters.sol:509 applies the **full delta** to each operator:

    if (deltaPositive) seb.operatorEthVUnits[operatorId] += deltaAbs;

The SPEC section 10 correctly describes this. Each operator tracks the full cluster vUnit deviation — this is needed because operator earnings are computed as `effectiveVUnits = operatorEthVUnits[id] + ethValidatorCount * VUNITS_PRECISION` and fees are charged at the cluster level using the full vUnits. Dividing by operatorCount would break the accounting.

The DIP-X proposal describes fees as scaling with total effective balance per cluster:

> **DIP-X proposal, "Accounting Changes" section:**
> "Fees are defined per 32 ETH of effective balance and scale with a cluster's total effective balance, rather than with validator count"

Each operator must track the full cluster-level deviation to correctly compute their share of fees from the cluster's total effective balance.

**Fix:** `operatorEthVUnits[opId] += (newVUnits - effectiveOldVUnits)`

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+
+---
+
+## Global Invariants (Must Always Hold)
+
+These invariants should be verified across all flows:
+
+1. **ETH conservation**: `contract.ETH_balance >= Σ(all active ETH cluster balances) + Σ(all operator ETH earnings) + staking_pool_balance`
+2. **SSV conservation**: `contract.SSV_balance >= Σ(all active SSV cluster balances) + Σ(all operator SSV earnings) + Σ(staked SSV)`
+3. **Validator count consistency**: `ethDaoValidatorCount == Σ(operator.ethValidatorCount)` across all operators (accounting for shared operators)
```
olegshmuelov: **Severity: HIGH** — mathematically wrong invariant, misleading for auditors/developers

This invariant is wrong. Operators are shared between clusters, so Σ(operator.ethValidatorCount) double-counts.

Example: 1 cluster with 2 validators using 4 operators → ethDaoValidatorCount = 2 but Σ(operator.ethValidatorCount) = 4 × 2 = 8.

The DIP-X proposal clarifies that the DAO tracks at the cluster level, not operator level:

> **DIP-X proposal, "New Clusters" section:**
> "ETH must be deposited upfront to fund the cluster's operational runway"

> **DIP-X proposal, "Accounting Changes" section:**
> "Here, total effective balance refers to the cumulative effective balance of all validators belonging to the cluster. All accounting is performed using this aggregated cluster-level value."

The correct invariant:

    ethDaoValidatorCount == Σ(cluster.validatorCount) across all active ETH clusters

The parenthetical "(accounting for shared operators)" acknowledges the complication but the formula doesn't handle it.

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+
+---
+
+### 1.3 Remove Validator
+
+**Caller:** Cluster owner
+
+#### Preconditions
+- Validator must exist and be owned by caller
+- Cluster must exist (ETH or SSV version)
```
olegshmuelov: **Severity: LOW** — documentation inaccuracy, no code impact

SSV clusters are NOT supported for removeValidator. The code at SSVValidators.sol:177 enforces VERSION_ETH only:

    ClusterLib.validateClusterVersion(version, VERSION_ETH);

Attempting to remove a validator from an SSV cluster reverts with IncorrectClusterVersion.

The DIP-X proposal explicitly lists this as a blocked operation for SSV clusters:

> **DIP-X proposal, "Existing Clusters (SSV-based)" section:**
> "This means that **adding new validators, removing existing validators, reactivating liquidated clusters** or depositing additional SSV to extend a cluster's runway **is no longer supported**."

This matches the SPEC section 1 ("Blocked operations: add validators, remove validators, reactivate, deposit SSV"). The FLOWS should match.

**Fix:** Change to "Cluster must exist as ETH cluster (VERSION_ETH)".

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+---
+
+### 1.4 Deposit ETH
+
+**Caller:** Anyone (on behalf of cluster owner)
+**Payable:** Yes
+
+#### Preconditions
+- Cluster must exist as ETH cluster (VERSION_ETH)
+- Cluster must be active
```
olegshmuelov: **Severity: MEDIUM** — spec/code mismatch; users/auditors may assume deposit is blocked for liquidated clusters

The deposit() code at SSVClusters.sol:190-205 does NOT check cluster.active. You can deposit ETH into a liquidated cluster. The function also does NOT do steps 1-2 ("Update operator snapshots", "Settle cluster fees"). It only does:

    cluster.balance += msg.value;
    s.ethClusters[hashedCluster] = cluster.hashClusterData();

This combines with the reactivation flow (line 300) which does cluster.balance += msg.value (not = msg.value as documented). If someone deposits to a liquidated cluster, then reactivation accumulates both amounts.

The DIP-X proposal does not restrict deposits to active clusters — it only says:

> **DIP-X proposal, "New Clusters" section:**
> "ETH must be deposited upfront to fund the cluster's operational runway"

Either:
1. Remove "Cluster must be active" precondition, remove steps 1-2, and document that deposit works on liquidated clusters too, OR
2. Add the active check and fee settlement to the code

docs/FLOWS.md
olegshmuelov, line 1: **Severity: LOW** — missing documentation for security-critical hook

Suggestion: Add a "5.6 cSSV Transfer" flow.

The DIP-X proposal explicitly specifies transfer behavior:

> **DIP-X proposal, "Rewards and Claiming" section:**
> "When cSSV is transferred, rewards accrued up to that point remain claimable by the original holder, while the new holder begins accruing rewards only from the moment they receive the cSSV."

The code implements this via CSSVToken._beforeTokenTransfer → SSVStaking.onCSSVTransfer(from, to, amount) which calls _syncFees() and _settle() for both sender and receiver before the transfer.

This hook is security-critical: without it, transferring cSSV right before a reward distribution would let the receiver claim rewards they didn't earn. Documenting the exact flow would help future auditors verify correctness.

olegshmuelov (changes requested):

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+```
+
+#### Postcondition Invariants
+- `operator.ethSnapshot.balance == previous_settled - amount`
+- `owner.balance == previous + amount`
+- `contract.balance == previous - amount`
+
+---
+
+### 4.8 Withdraw Operator Earnings (SSV)
```
RaekwonIII, line 729: As discussed: `OperatorWithdrawnSSV` (or equivalent),  SSV-specific event

docs/FLOWS.md
```diff
@@ -0,0 +1,1005 @@
...
+3. Reset operator state via `_resetOperatorState`: zeros `ethSnapshot`, `snapshot`, `ethFee`, `fee`, `ethValidatorCount`, `validatorCount`
+4. **`operator.owner` is intentionally preserved** — allows off-chain systems (explorer, `getOperatorById`) to query the original owner after removal
+5. Withdraw all SSV earnings to owner (if any)
+6. Withdraw all ETH earnings to owner (if any)
+7. Delete whitelist mapping
+8. Delete fee change request (if any)
+
+#### Events
+```solidity
+if (ssvEarnings > 0) emit OperatorWithdrawn(owner, operatorId, ssvEarnings);
```
RaekwonIII: As discussed on Discord, let's please create `OperatorWithdrawnSSV` event (or similar) for SSV-specific earnings.

docs/SPEC.md
```diff
@@ -0,0 +1,836 @@
...
+
+### Operator Events
+
+```solidity
+event OperatorAdded(uint64 indexed operatorId, address indexed owner, bytes publicKey, uint256 fee);
+event OperatorRemoved(uint64 indexed operatorId);
+event OperatorFeeDeclared(address indexed owner, uint64 indexed operatorId, uint256 blockNumber, uint256 fee);
+event OperatorFeeDeclarationCancelled(address indexed owner, uint64 indexed operatorId);
+event OperatorFeeExecuted(address indexed owner, uint64 indexed operatorId, uint256 blockNumber, uint256 fee);
+event OperatorWithdrawn(address indexed owner, uint64 indexed operatorId, uint256 value);
```
RaekwonIII, line 651: `event OperatorWithdrawnSSV` to be added 🙏

docs/SPEC.md
```diff
@@ -0,0 +1,836 @@
...
+| `minimumBlocksBeforeLiquidation` | 50,190 blocks (~7 days) | `updateLiquidationThresholdPeriod(uint64)` |
+| `operatorMaxFee` | TBD | `updateMaximumOperatorFee(uint256)` |
+| `minimumOperatorEthFee` | TBD | `updateMinimumOperatorEthFee(uint256)` |
+
+### SSV Cluster Parameters (Legacy)
+
+| Parameter | Current Value | Proposed Value |
+|---|---|---|
+| `minimumLiquidationCollateralSSV` | 1.53 SSV | 0.883 SSV |
+| `minimumBlocksBeforeLiquidationSSV` | 100,380 (~14 days) | 100,380 (~14 days) |
```
RaekwonIII: We still have the SSV-equivalents for SSV legacy clusters, no?
e.g. `networkFeeSSV`, `minimumLiquidationCollateralSSV`, `operatorMaxFeeSSV`, `minimumOperatorEthFeeSSV` and so forth, no?

RaekwonIII (changes requested):
Requested changes mostly because of the new event. Rest are just comments/curiosity/feedback

docs/SPEC.md
RaekwonIII: Means the smallest increment is an integer, i.e. 33 ETH, not 33.4 ETH.
And afaik this is how the beacon chain represents it as well.

RaekwonIII:
@liorrutenberg watch out, your commits appear to not be **verified**. AFAIK in our repos the policy stops PRs from being merged if this is the case.

docs/SPEC.md
mtabasco: Correct, I rephrased the statement to clarify

docs/FLOWS.md
mtabasco: Correct, I updated the statement to clarify.

docs/FLOWS.md
mtabasco: No. I added a note to this step to clarify.

docs/FLOWS.md
mtabasco: No. Same as `executeOperatorFee`.

docs/FLOWS.md
mtabasco: yeah! I added it

docs/FLOWS.md
mtabasco: @arielssv This check is not needed as it is implemented by construction. cSSV tokens are minted after the SSV tokens are transferred, so `cSSV.totalSupply` is always <= SSV held by the SSVNetwork contract <= `SSV.totalSupply`.

I added a note clarifying it.

docs/SPEC.md
mtabasco: Yes, section updated

docs/SPEC.md
mtabasco: @arielssv The choice of `timestamp` is to facilitate a better UX, showing users (and UI) the precise moment when the tokens are unlocked. It is the same pattern as the declare -> execute operator fee change.
Any chance the proposal can be changed to use `timestamp`? Thanks

docs/SPEC.md
mtabasco: This is a discrepancy that should be addressed @arielssv

docs/SPEC.md
arielssv: yes will change proposal

docs/FLOWS.md
arielssv: you suggest that if there are no bugs cSSV.total supply < SSV.total supply. i suggest to enforce it regardless. but up to you

CLAUDE.md
```diff
@@ -0,0 +1,261 @@
...
+## Build & Test Commands
+
+```bash
+npm install                    # Install dependencies
+npx hardhat compile            # Compile contracts
+npm run test                   # Run all tests
+npm run test:unit              # Run unit tests only
+npm run test:integration       # Run integration tests only
+npm run test-forked            # Run fork tests (requires fork config)
+npx hardhat coverage           # Generate coverage report
```
yurii-ssv: hardhat-based coverage was removed in hh v3, this command will not work. we have a just recipe to check coverage (just coverage)

It runs hh tests with coverage flag and uses gnhtm to generate output from lcov later 

coverage:
    COVERAGE=true npx hardhat test --coverage
    genhtml coverage/lcov.info -o coverage/html

CLAUDE.md
```diff
@@ -0,0 +1,261 @@
...
+
+## Build & Test Commands
+
+```bash
+npm install                    # Install dependencies
+npx hardhat compile            # Compile contracts
+npm run test                   # Run all tests
+npm run test:unit              # Run unit tests only
+npm run test:integration       # Run integration tests only
+npm run test-forked            # Run fork tests (requires fork config)
```
yurii-ssv: this command will not work properly, current fork setup is made via hardhat suite and activated via env value, I suggest to use just to run fork tests only

test-forked:
    NO_GAS_ENFORCE=true RUN_FORK=true npx hardhat test $(find test/test-forked -name "*.test.ts" | xargs)

CLAUDE.md
```diff
@@ -0,0 +1,261 @@
...
+npx hardhat coverage           # Generate coverage report
+```
```
yurii-ssv, line 26: Addressed this table and issue in [PR #446 ](https://github.com/ssvlabs/ssv-network/pull/446), I think needs to be merged before merging this branch

CLAUDE.md
```diff
@@ -0,0 +1,261 @@
...
+| Parameter | Value | Update Function |
+|---|---|---|
+| ethNetworkFee | 0.000000003550929823 ETH/block (~0.00928 ETH/year) | `updateNetworkFee(uint256)` |
+| minimumLiquidationCollateral | 0.00094 ETH | `updateMinimumLiquidationCollateral(uint256)` |
+| minimumBlocksBeforeLiquidation | 50190 (~7 days) | `updateLiquidationThresholdPeriod(uint64)` |
+| defaultOperatorETHFee | 0.000000001775464912 ETH/block (~0.00464 ETH/year) | Hardcoded in contract |
+| cooldownDuration | 50120 blocks (~7 days) | `setUnstakeCooldownDuration(uint64)` |
+| quorumBps | 7500 (75%) | `setQuorumBps(uint16)` |
+| Oracle set | 4 oracles, 3-of-4 threshold | `replaceOracle(uint32, address)` |
+
```
yurii-ssv, line 130: Maybe better to add real values we intend to use too? I mean packable versions

CLAUDE.md
```diff
@@ -0,0 +1,261 @@
...
+## Governance Parameters (DIP-X Proposed Values)
+
+| Parameter | Value | Update Function |
+|---|---|---|
+| ethNetworkFee | 0.000000003550929823 ETH/block (~0.00928 ETH/year) | `updateNetworkFee(uint256)` |
+| minimumLiquidationCollateral | 0.00094 ETH | `updateMinimumLiquidationCollateral(uint256)` |
+| minimumBlocksBeforeLiquidation | 50190 (~7 days) | `updateLiquidationThresholdPeriod(uint64)` |
+| defaultOperatorETHFee | 0.000000001775464912 ETH/block (~0.00464 ETH/year) | Hardcoded in contract |
+| cooldownDuration | 50120 blocks (~7 days) | `setUnstakeCooldownDuration(uint64)` |
```
yurii-ssv: We use seconds now, maybe worth to mention it here

CLAUDE.md
```diff
@@ -0,0 +1,261 @@
...
+│   └── CSSVToken.sol                 # cSSV receipt token (mint/burn by SSVStaking only)
+├── whitelisting/BasicWhitelisting.sol
+└── upgrades/stage/hoodi/            # Upgrade initializer (reinitializer(3))
+scripts/                              # Deployment & upgrade scripts (TypeScript)
+test/
+├── unit/                             # Per-module unit tests
+├── integration/                      # Full integration tests
+├── sanity/                           # Sanity/regression tests
+├── echidna/                          # Foundry-based fuzzing
+├── test-forked/                      # Fork tests against v2.0.0
```
yurii-ssv: against v1.2.0 **

yurii-ssv (changes requested):

venimir-ssv (approved):

docs/FLOWS.md
mtabasco: I added a new task to create specific tests to probe this invariant:
`cSSV.totalSupply() <= SSV.balanceOf(SSVStaking)`.
No need to enforce the check via code.

docs/SPEC.md
mtabasco: Confirmed with @arielssv: we should keep the current `burn-then-wait` behaviour.

docs/FLOWS.md
mtabasco: Added a note on the Deposit ETH section:
> **Note — deposits allowed on liquidated clusters:** `deposit` does not require the cluster to be active. Depositing to a liquidated cluster, and later reactivating it, will accumulate both the deposit and the reactivation amount.

docs/FLOWS.md
mtabasco: Done

docs/SPEC.md
mtabasco: updated

docs/SPEC.md
mtabasco: done

docs/FLOWS.md
mtabasco: done

docs/FLOWS.md
mtabasco: We will enhance tests to assert solvency invariants.

docs/FLOWS.md
mtabasco: updated

docs/FLOWS.md
mtabasco: updated

docs/FLOWS.md
mtabasco: updated description

CLAUDE.md
mtabasco: updated in [PR #446 ](https://github.com/ssvlabs/ssv-network/pull/446)

CLAUDE.md
mtabasco: updated in https://github.com/ssvlabs/ssv-network/pull/446

CLAUDE.md
mtabasco: done

CLAUDE.md
mtabasco: updated

CLAUDE.md
mtabasco: updated

docs/SPEC.md
mtabasco: Yes, updated

docs/SPEC.md
mtabasco: added specific comments to explain:
- What happens when the final oracle votes but the proposal still doesn't reach quorum?
- How is this state cleaned up?
- Can the same blockNum be re-proposed with a different root?

docs/FLOWS.md
mtabasco: Added a task to address the change or not.

docs/FLOWS.md
mtabasco: Yes, it is referenced. For pre-upgrade declarations, when executing them, we revert the transaction. The user needs to declare it again (ETH fee).

docs/SPEC.md
mtabasco: updated

mtabasco (approved):

CLAUDE.md
mtabasco: done

Amir-SSVLabs (approved):
Safe to merge once:
1. The team confirms the mainnetRpcUrl fork config is intended and that MAINNET_ETH_NODE_URL is documented for fork tests.
2. Any critical links to the removed docs are updated to the new SPEC/FLOWS/CLAUDE structure.
[Both approved by Marco]

PR #438, opened 2026-02-19
Refactor deployment scripts

mtabasco:
### Summary

Replaces the previous ad-hoc config file pattern with a structured, env-based deployment system. Each environment (`hoodi-stage`, `hoodi-prod`, etc.) has a dedicated folder under `deployments/` containing its `config.json` and all result files. Scripts are invoked via `just` recipes with a single `env` parameter. Adds mandatory version pre-flight checks and versioned result file history.

---

#### `just` recipes — single `env` parameter

All deployment and upgrade operations now accept one `env` argument that resolves config and output paths automatically:

```sh
just upgrade hoodi-stage        # upgrades using deployments/hoodi-stage/config.json
just upgrade hoodi-prod         # upgrades using deployments/hoodi-prod/config.json
just deploy hoodi-stage         # deploys impls/modules, writes result to deployments/hoodi-stage/
just generate-safe-batch hoodi-prod  # generates multisig batch from env config + deploy-result
```

The script reads `deployments/<env>/config.json` and writes all results back to `deployments/<env>/`.

Both `deploy-fresh` and `upgrade` recipes now run `npx hardhat compile --force` before executing to ensure artifacts are always current.

#### Versioned result files with symlink history

Every run produces a version-stamped result file alongside a fixed-name symlink pointing to the latest:

```
deployments/hoodi-stage/
  config.json
  upgrade-result.v2.0.1.json   ← canonical record for this run
  upgrade-result.json          ← symlink → upgrade-result.v2.0.1.json (latest)
```

Re-running after a future upgrade produces `upgrade-result.v2.0.2.json` and updates the symlink — the previous versioned file is preserved as a permanent audit trail. The same pattern applies to `deploy-result.<version>.json`.

#### Version pre-flight checks (hard errors before any deployment)

Two checks run before any contract is deployed or upgraded, aborting immediately on mismatch:

1. **`currentVersion`** — reads `getVersion()` from the live on-chain proxy and asserts it matches `config.currentVersion`. Prevents running an upgrade script against the wrong proxy or environment.

2. **`targetVersion`** — parses `getVersion()` directly from `contracts/libraries/CoreLib.sol` source using a regex. Asserts the local source matches `config.targetVersion`. No deployment, no gas, no RPC call needed. Prevents deploying a mismatched implementation.

```
[PRE-FLIGHT] currentVersion = v2.0.0 ✓
[PRE-FLIGHT] targetVersion  = v2.0.1 ✓
```

If either check fails, the script exits with a descriptive error before touching any contract.

#### `skipInitializer` flag

Added `skipInitializer?: boolean` to `UpgradeConfig`. When `true`, the proxy upgrade uses `upgradeTo` instead of `upgradeToAndCall`, skipping the initializer call. Used for patch upgrades where the initializer was already executed in a prior version upgrade.

```json
{
  "skipInitializer": true
}
```

#### `config.json` schema additions

```json
{
  "currentVersion": "v2.0.0",
  "targetVersion": "v2.0.1",
  "skipInitializer": true
}
```

| Field | Required | Description |
|---|---|---|
| `currentVersion` | Yes | Version string expected from the live on-chain proxy. Pre-flight aborts if mismatch. |
| `targetVersion` | Yes | Version string expected from the new implementation source. Pre-flight aborts if mismatch. Also used as the result file suffix. |
| `skipInitializer` | No | If `true`, uses `upgradeTo` instead of `upgradeToAndCall`. Default: `false`. |

#### Code refactoring

- `scripts/common/config.ts` — extracted all config types, path resolvers, oracle helpers, protocol param resolution, and symlink utilities into a single shared module. Scripts no longer duplicate this logic.
- `scripts/upgrade.ts` — rewritten to use the shared config module; removed inline path logic, inline oracle normalization, and inline param parsing.
- `scripts/deploy.ts` — new script for deploying implementations and modules only (no proxy upgrade), producing a `deploy-result.<version>.json` consumed by `generate-safe-batch`.
- `scripts/generate-safe-batch.ts` — new script that reads `config.json` + `deploy-result.json` and produces a Safe Transaction Builder JSON batch for multisig execution.
- Removed `scripts/upgrade-fork.ts` (superseded by `upgrade.ts` with `--fork` flag).
- Removed `deployments/hoodi-upgrade.config.json` (superseded by `deployments/hoodi-stage/config.json`).
- Added `deployments/template-config.json` as a reference for new environments.
- Updated `deployments/README.md` to reflect the new structure and config schema.


mtabasco:
replaced by https://github.com/ssvlabs/ssv-network/pull/439

PR #457, opened 2026-02-25
feat: cleanup and correct e2e test scenarios

yurii-ssv:


test/e2e/helpers/invariant-checker.ts
```diff
@@ -1,119 +1,89 @@
...
   operatorEarnings: bigint[],
-  stakingPool: bigint,
-  daoETHEarnings: bigint,
+  networkTotalEarnings: bigint,
 ): Promise<void> {
   const contractBalance = await provider.getBalance(contractAddress);
 
   const totalClusters = clusterBalances.reduce((sum, b) => sum + b, 0n);
   const totalOperators = operatorEarnings.reduce((sum, b) => sum + b, 0n);
-  const totalAccounted = totalClusters + totalOperators + stakingPool + daoETHEarnings;
```
mtabasco, line 35: @yurii-ssv This was incorrect. We can not sum stakingPool + daoETHEarnings. 
`networkTotalEarnings()` already includes all current DAO earnings, which are the source of staking rewards. The `stakingEthPoolBalance` is an accounting pointer that tracks how much of `ethDaoBalance` has been distributed to stakers, but it's NOT a separate balance in the contract.

The bug is masked because `stakingPool` is always `0` in all test calls.

I updated the formulas in `SPEC` and `FLOWS`

test/e2e/helpers/invariant-checker.ts
```diff
@@ -1,119 +1,89 @@
...
  *
- * @param views - SSVNetworkViews contract
- * @param operatorIds - array of operator IDs to check
+ * IMPORTANT: This requires test-side tracking of all clusters because the contract
+ * does not expose an iterator over clusters. Pass all clusters created during the test.
+ *
+ * NOTE: Σ(operator.ethValidatorCount) is NOT equivalent because operators are shared
+ * across clusters and would overcount validators (see SPEC.md §11.3).
  */
 export async function checkValidatorCountConsistency(
```
mtabasco, line 48: The previous function was not used and was semantically wrong. The correct formula is:
`ethDaoValidatorCount == Σ(cluster.validatorCount) across all active ETH clusters`

I created a couple of tests to check it, although it would be great to test this invariant with echidna.

mtabasco (approved):

PR #463, opened 2026-02-25
TEST-4 enrich tests

mtabasco:
Before (Integration Tests):
❌ Used greaterThan, lessThan, greaterThanOrEqual (11 violations)
❌ Used "tolerance" pattern allowing 100 wei rounding error
❌ Didn't track block numbers for precise calculations
After (Integration Tests):
✅ All assertions use exact .to.equal() with calculated expected values
✅ Block numbers tracked precisely for fee calculations
✅ Every balance assertion references SPEC.md formulas

Added 3 New Edge Case Tests:
test/unit/SSVClusters/updateClusterBalance.test.ts
Test 1: Multi-Validator Liquidated Cluster (Lines 463-527)
Test 2: EB Decrease on Liquidated Cluster (Lines 529-602)
Test 3: Implicit→Explicit EB Transition (Lines 604-653)

### DIFF
--- a/test/common/helpers.ts
+++ b/test/common/helpers.ts
@@ -168,7 +168,8 @@ export async function registerDefaultCluster(
 ): Promise<{
   cluster: Cluster,
   validatorKey: string,
-  operatorIds: number[]
+  operatorIds: number[],
+  receiptRegister: any
 }> {
   const validatorKey = makePublicKey(1);
   const operatorIds = await registerOperators(network, operatorOwner, 4);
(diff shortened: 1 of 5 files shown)


test/common/helpers.ts
```diff
@@ -195,7 +197,7 @@ export async function registerDefaultCluster(
   );
 
   return {
-    cluster, validatorKey, operatorIds
+    cluster, validatorKey, operatorIds, receiptRegister
```
mtabasco, line 200: @yurii-ssv do you think adding tx receipt here is useful? I only use it to get the block number in one test...

yurii-ssv: Yes, why not. Can be very helpful when we will start addressing #435

yurii-ssv (approved):

PR #479, opened 2026-02-27
TEST-26	Zero-validator cluster operations

yurii-ssv:


venimir-ssv (approved):

gabriel-ssvlabs:
---                                                                                                                                                                                                                                                                                    
  PR #479 — TEST-26: Zero-Validator Cluster Operations
                                                                                                                                                                                                                                                                                         
  Branch: fix/test-26
  Scope: deposit.test.ts (+29 lines), withdraw.test.ts (+29 lines), updateClusterBalance.test.ts (+95 lines), MAINNET-READINESS.md (updated)

  ---
  What this PR adds

  Four sub-tasks proving that zero-validator ETH clusters behave correctly across three modules:

  1. Deposit — balance increases by the exact deposit amount (no fees charged during the gap)
  2. Withdraw — full balance can be withdrawn; cluster stays active with balance == 0
  3. EB update (basic) — updateClusterBalance(EB=0, proof=[]) on a zero-validator cluster emits the event, reports getClusterVUnits == 0, leaves DAO state unchanged
  4. EB update (full state) — a cluster that previously had an explicit EB of 64 ETH, then had its last validator removed (triggering ebSnapshot.vUnits = 0 cleanup in removeValidator), can receive a second updateClusterBalance(EB=0) call without corrupting any state

  ---
  1. Spec / Flows Deviations

  None found.

  The tests rely on four contract-level behaviours that needed verification:
<img width="1330" height="189" alt="image" src="https://github.com/user-attachments/assets/dbd409bf-e39b-4832-8b5a-1494c63824d4" />


  The most subtle one is the single-leaf Merkle proof. The tests call mockSetEBRoot(blockNum, getEBRoot(clusterId, 0)) where getEBRoot computes keccak256(keccak256(abi.encode(clusterId, 0))) and stores that hash directly as the committed root. Passing an empty proof [] then
  satisfies MerkleProof.verify because a single-node tree has no sibling hashes — root equals the leaf. This is valid and matches the OZ implementation.

  ---
  2. Acceptance Criteria

  
<img width="1720" height="191" alt="image" src="https://github.com/user-attachments/assets/b38e2957-2a6a-4855-aa90-18119689d8e1" />


  All four acceptance criteria are fully met.

  ---

  Sub-task 1 (deposit): The test elegantly captures the cluster balance at the moment the last validator is removed, then advances 100 blocks with no validators registered. By asserting clusterAfterDeposit.balance == balanceAtRemoval + depositAmount it implicitly proves that
  zero-validator clusters do not accrue any fees during the wait — because if they did, the pre-deposit balance would have drifted below balanceAtRemoval. The fixture uses a non-zero operator fee (10_000_000_000n) so a mis-implementation would be caught. Correct.

  Sub-task 2 (withdraw): Uses deploySSVClustersWithLowFeesFixture plus mockEthNetworkFee(100_000n) to keep fees deterministic. After removing the validator and mining 100 blocks, it withdraws the full remaining balance and checks balance == 0 and active == true. This verifies that
   withdraw does not deactivate a cluster and that the accounting settles to exactly zero. Correct.

  Sub-task 3 (updateClusterBalance basic): Uses the standard fixture. Registers and removes a validator, then calls updateClusterBalance with effectiveBalance=0 and proof=[]. Asserts that ClusterBalanceUpdated is emitted with effectiveBalance=0, that getClusterVUnits returns zero,
   and that daoTotalEthVUnits is unchanged. The DAO vUnits check is the most important invariant here — it confirms no phantom vUnits were credited or debited. Correct.

  Sub-task 4 (full state assertion): This is the richest test. The state machine it exercises:

  1. Register one validator (implicit EB, ebSnapshot.vUnits = 0)
  2. updateClusterBalance(blockNum=1, EB=64) — sets ebSnapshot.vUnits = 20000 (ceiling of 64×10000/32), updates operator and DAO deviations
  3. removeValidator — validatorCount drops to 0; the cleanup path in SSVValidators.sol reduces ebSnapshot.vUnits by VUNITS_PRECISION, sees validatorCount == 0, clears operator deviations and DAO deviation, then zeroes ebSnapshot.vUnits
  4. updateClusterBalance(blockNum=2, EB=0) — ebToVUnits(0) = 0, effectiveOldVUnits = ebSnapshot.vUnits = 0, no deviation update triggered, EB snapshot written as 0, liquidation guard returns immediately (validatorCount = 0)

  The mocks (minimumLiquidationCollateral=0, ethNetworkFee=0) are correctly motivated: they prevent the updateClusterBalance(EB=64) call from auto-liquidating or from complicating fee arithmetic, keeping the state assertions clean. The six named criteria — EB limits pass,
  getClusterVUnits == 0, clusterEB.vUnits == 0, no operator/DAO vUnit changes, cluster still active, event emitted with effectiveBalance=0 — cover all meaningful observable state. Correct.

  ---
  Minor observations

  Asymmetry between deposit and withdraw tests: The deposit test explicitly proves no fees accrued during the 100-block zero-validator gap (via the exact balanceAtRemoval + depositAmount check). The withdraw test does not assert this independently — it only asserts the result of
  withdrawing. This is acceptable because AC1 already covers fee-free accrual; the withdraw AC is specifically about mechanics (full withdrawal, cluster stays active). But if someone later broke fee accounting for zero-validator clusters in a way that only affected the withdraw
  path, this test would not catch it. Low risk given the deposit test already guards the same code path.

  ethNetworkFee = 0 in sub-task 4: The mock silences network fee accrual during updateClusterBalance(EB=64). Without it, operator fees could still run against the 10 ETH deposit over the block range, but minimumLiquidationCollateral=0 already prevents auto-liquidation. Setting
  ethNetworkFee=0 is the tidiest choice here — it avoids needing to calculate expected fee deductions in the assertions.

  ---
  Verdict

  Approve. All four acceptance criteria are satisfied. The contract behaviours underpinning the tests (ebToVUnits(0), _verifyEBLimits with zero inputs, isLiquidatableWithEB early return, removeValidator EB cleanup, single-leaf Merkle proof) were each individually verified against
  the source and are correct. No deviations from docs/SPEC.md or docs/FLOWS.md. The sub-task 4 state machine trace is thorough and covers the most non-obvious interaction between the EB oracle and the validator remove lifecycle.

PR #480, opened 2026-02-27
TEST-27	Operator at max validator limit

yurii-ssv:


venimir-ssv (approved):

gabriel-ssvlabs:
---                                                                                                                                                                                                                                                                                    
  PR #480 — TEST-27: Operator at Max Validator Limit                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                         
  Branch: fix/test-27                                                                                                                                                                                                                                                                    
  Scope: test/unit/SSVValidator/registerValidator.test.ts (+44 lines), MAINNET-READINESS.md (status update)

  ---
  What this PR adds

  Two new tests appended to registerValidator.test.ts:

  1. Sub-task 1 — Register a validator that pushes an operator over its limit → verify ExceedValidatorLimitWithData revert
  2. Sub-task 2 — Remove one validator to drop back below the limit, then re-register → verify success

  Both tests use mockValidatorsPerOperatorLimit(5) to set a practical test boundary instead of requiring 3000 on-chain registrations.

  ---
  1. Spec / Flows Deviations

  None found. The contract behaviour being tested is:

  // OperatorLib.sol:233
  if ((operator.ethValidatorCount += deltaValidatorCount) > sp.validatorsPerOperatorLimit) {
      revert ISSVNetworkCore.ExceedValidatorLimitWithData(operatorId);
  }

  This check runs inside updateClusterOperatorsOnRegistration for both single and bulk registration. For bulk, deltaValidatorCount = publicKeys.length (the entire batch is counted at once), so registering 5 validators in one bulk call atomically moves each operator's count from 0
  → 5. 5 > 5 is false, so the limit is exactly reached, not exceeded. That boundary condition is what the test exercises. Matches SPEC §8 ExceedValidatorLimitWithData error code and the operator flow in FLOWS §4.1 preconditions.

  ---
  2. Acceptance Criteria
<img width="1663" height="124" alt="image" src="https://github.com/user-attachments/assets/c8dc0e3a-4045-4152-87a6-6a6ca72b93af" />

 
  Both acceptance criteria are met.

  ---

  parseClusterFromEvent on a bulk receipt: This deserves a close look because bulkRegisterValidator emits one ValidatorAdded event per validator, and parseClusterFromEvent returns receipt.eventsByName[eventName][0] — the first event. For single-validator registration that is
  clearly correct; for bulk, the question is whether all events carry the same cluster struct or sequential ones.

  Reading _bulkRegisterValidator in SSVValidators.sol:115: cluster.updateClusterOnRegistration(...) is called once before the event loop, incrementing cluster.validatorCount by the full validatorsLength and storing the final hash. The subsequent emit ValidatorAdded loop emits the
  same cluster memory variable for every key. All five events therefore carry an identical cluster with validatorCount = 5n. parseClusterFromEvent returning [0] is correct here. clusterAtLimit has the right state to pass validateHashedCluster in the next call. ✅

  Sub-task 1 (revert):
  - Bulk-registers 5 validators: each operator.ethValidatorCount goes 0 → 5; check 5 > 5 = false ✅
  - registerValidator with delta=1: each operator goes 5 → 6; check 6 > 5 = true → revert with ExceedValidatorLimitWithData(operatorIds[0]) ✅
  - .withArgs(operatorIds[0]) is correct — the first operator in the sorted list is the first to trip the check in the loop ✅

  Sub-task 2 (re-register after remove):
  - removeValidator(publicKeys[0], ...) decrements each operator from 5 → 4 ✅
  - parseClusterFromEvent on the remove receipt correctly gets the cluster post-removal (single event, no ambiguity) ✅
  - registerValidator(makePublicKey(6), ...) — uses a fresh key (not key 1 which was removed, though that would also work since the record is deleted); each operator goes 4 → 5; check 5 > 5 = false → succeeds ✅
  - Asserts emit ValidatorAdded ✅

  Mock setup: ssvValidatorsHarnessFixture initialises the limit at 3000 (line 88 of fixtures.ts). The tests immediately override it to 5. This is the correct pattern — testing the exact boundary without requiring 3000 actual registrations.

  ---
  Minor observations

  Duplicate setup block: Both tests replicate the same 8-line bulk registration setup (mock limit, makePublicKeys, bulkRegisterValidator, parseClusterFromEvent). The second test could load a pre-built inner fixture to avoid the duplication, but at only 2 tests this is a low-signal
   issue.

  Weak assertion in sub-task 2: The re-register test only checks emit ValidatorAdded without verifying the cluster's validatorCount or the operator's ethValidatorCount via the harness view function. The AC says "verify succeeds" so this is technically sufficient, but adding
  .withArgs(owner, operatorIds, makePublicKey(6), DEFAULT_SHARES, ...) or a follow-up getOperatorEthValidatorCount check would make the test self-documenting and catch regressions more precisely.

  { value: DEFAULT_ETH_REGISTER_VALUE } in sub-task 2 re-register call: The cluster already holds 10 ETH from the original bulk deposit (and operator fees are 0), so the re-registration would succeed with { value: 0 } as well. Passing 10 ETH is harmless but slightly misleading — a
   reader might think the additional deposit is required.

  ---
  Verdict

  Approve. Both acceptance criteria are met. The contract-level limit check (> validatorsPerOperatorLimit), the operator ID in the error args, the cluster state after bulk registration, and the remove-then-re-register path are all exercised correctly. The parseClusterFromEvent
  edge case (bulk events all sharing the same cluster struct) is safe. The minor observations are cosmetic and do not affect test correctness.

mtabasco (approved):
@yurii-ssv please resolve conflicts on MAINNET-READINESS and good to go

PR #482, opened 2026-02-27
Test 10 fee change eb burn rate

venimir-ssv:


yurii-ssv (approved):

gabriel-ssvlabs:
---                                                                                                                                                                                                                                                                                    
  PR #482 — TEST-10: Operator Fee Change + EB Burn Rate Interaction                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                         
  Branch: test-10-fee-change-eb-burn-rate                                                                                                                                                                                                                                                
  Scope: test/unit/SSVClusters/feeChangeEBInteraction.test.ts (+323 lines, new file), MAINNET-READINESS.md (status update)

  ---
  What this PR adds

  A new test file with three integration-style tests using ssvNetworkFullFixture (real network, not a harness), a local settleClusterAtEB helper that commits oracle roots and calls updateClusterBalance, and a local registerAndSetEb64Cluster helper that wires up a 4-operator
  cluster with explicit EB=64:

  1. AC1 — Declare + execute a 2× fee increase on all 4 operators; measure deduction over equal windows before and after → assert deductionAfter == deductionBefore * 2
  2. AC2 — Start at 2× fee, reduceOperatorFee to 1×; measure equal windows → assert highFeeDeduction == lowFeeDeduction * 2
  3. AC3 — Change fee on ONE operator only; assert exact deduction from first principles using the SPEC §10 cumulative index formula

  ---
  1. Spec / Flows Deviations

  None found. The tests correctly model the fee settlement rule from SPEC §10:

  When an operator fee changes, updateSnapshotSt settles earnings at the OLD rate up to the current block; the new fee applies only to subsequent blocks.

  The key contract paths verified:

  - executeOperatorFee calls OperatorLib.updateSnapshotSt(operator, operatorId) before writing operator.ethFee = newFee — old rate frozen at executeBlock ✅
  - reduceOperatorFee reads operator into memory, calls operator.updateSnapshot(operatorId), then writes the snapshot + new fee back to storage — same settlement guarantee ✅
  - updateBalanceWithEB formula: idxOp = newIndex - cluster.index, usage = (idxOp * vUnits / VUNITS_PRECISION) * ETH_DEDUCTED_DIGITS — with vUnits = 20000 for EB=64 and VUNITS_PRECISION = 10000, the factor is exactly 2, no rounding error ✅
  - Network fee zeroed via updateNetworkFee(0n) in the fixture to isolate operator fee behavior. currentNetworkFeeIndex stays constant, so idxNet = 0 and network fees contribute nothing to the measurements ✅

  ---
  2. Acceptance Criteria

 
<img width="1516" height="164" alt="image" src="https://github.com/user-attachments/assets/db7aec60-5ce9-43a6-a3b2-6a062f26aba5" />


  ---

  Measurement symmetry (Tests 1 and 2):

  Both tests measure deductions over two windows of equal block length. Each settleClusterAtEB call internally executes the same sequence: mine(1) + 3 commitRoot txs + 1 updateClusterBalance tx = 5 blocks. Combined with mine(windowBlocks = 120), each measurement window spans
  exactly 125 blocks. Since both windows are the same length and the EB stays at 64 throughout, a 2× fee change produces exactly 2× deduction. ✅

  Boundary formula verification (Test 3):

  Only operatorIds[0] gets the fee increase. The formula:

  const expectedIndexDelta =
    oldSpanBlocks * (oldPackedFee * 4n) +
    newSpanBlocks * (oldPackedFee * 3n + newPackedFee);

  Traces correctly through the contract mechanics:

  - During oldSpanBlocks = executeBlock − settledBlock: all 4 operators still at oldPackedFee. The contract's updateClusterOperators accumulates 4 × oldPackedFee × blockDiff into the cumulative index.
  - executeOperatorFee calls updateSnapshotSt for operator 0 at executeBlock (old fee locked in), then sets ethFee = newPackedFee. Operators 1–3 are untouched.
  - During newSpanBlocks = finalBlock − executeBlock: updateClusterOperators in the final updateClusterBalance call advances operator 0 from executeBlock at newPackedFee, and operators 1–3 from settledBlock at oldPackedFee. The index delta contribution = 3 × oldPackedFee ×
  (finalBlock − settledBlock) + newPackedFee × (finalBlock − executeBlock), which decomposes to 3 × oldPackedFee × oldSpanBlocks + 3 × oldPackedFee × newSpanBlocks + newPackedFee × newSpanBlocks → matches the formula. ✅

  The clusterAtEb64 cluster struct (from initial EB=64 settlement) is safely reused in the final settleClusterAtEB call because no cluster-mutating operations happened between settledBlock and the final settlement — only operator fee changes and oracle commits, neither of which
  updates the stored cluster hash. ✅

  Oracle setup:

  The fixture correctly initialises 4 oracles with replaceOracle(1..4, ...) and stakes STAKE_AMOUNT SSV to produce cSSV supply. Oracle weight per slot = totalCSSVSupply / 4. Three oracles × weight = 75% = exactly quorumBps = 7500. Quorum is met on the third commit. ✅

  DECLARE_OPERATOR_FEE_PERIOD unit mismatch — no bug:

  DECLARE_OPERATOR_FEE_PERIOD = 604800n is defined in seconds (7 days). The tests call mine(DECLARE_OPERATOR_FEE_PERIOD + 1n) which mines 604,801 blocks. Hardhat advances block.timestamp by 1 second per mined block, so the fee window advances by 604,801 seconds — just past the
  approvalBeginTime. The executeOperatorFee timestamp check passes. ✅ This works but using a block-mining constant to advance a timestamp-based window is confusing. A comment explaining the Hardhat timestamp-per-block behavior would help future readers.

  Packed fee arithmetic:

  oldPackedFee = MINIMAL_OPERATOR_ETH_FEE / ETH_DEDUCTED_DIGITS = 1_770_000_000 / 100_000 = 17_700. Division is exact (no remainder). newPackedFee = 35_400. Both match the packed values the contract stores via PackedETHLib.pack(fee). ✅

  ---
  Minor observations

  EB state not explicitly asserted: Test 3 assumes vUnits = 20_000 (EB=64) throughout, but never calls getClusterVUnits to verify the EB is correctly set. The math fails silently if EB isn't applied. Other tests in the codebase cover EB setting, so the risk is low, but an
  assertion would make test 3 self-contained.

  settleClusterAtEB adds unaccounted blocks to the before-increase window in test 1: deductionBeforeIncrease is measured from clusterAtEb64.balance (from registerAndSetEb64Cluster) to beforeIncrease.cluster.balance. The start block is the tail of registerAndSetEb64Cluster, not a
  clean state. This is intentional by design — both windows start and end with a settleClusterAtEB call of the same structure, so the comparison is valid. ✅

  Setup duplication between tests 1 and 2: Both tests duplicate the mine(windowBlocks) → first settle → fee change → afterChangeSettle → mine(windowBlocks) → second settle pattern. The helper registerAndSetEb64Cluster reduces some of this, but an inner fixture for the "cluster at
  high fee, first window measured" state could further reduce repetition.

  ---
  Verdict

  Approve. All three acceptance criteria are met. The formula in test 3 is mathematically correct and faithfully models the SPEC §10 fee settlement rule for a mixed-fee-rate cluster. The symmetry argument for tests 1 and 2 holds precisely. The DECLARE_OPERATOR_FEE_PERIOD
  block-mining approach is a common Hardhat pattern and works correctly here. No spec or flow deviations.

mtabasco (approved):
@venimir-ssv please resolve conflicts and merge

PR #501, opened 2026-03-03
Complete FUZZ-5 ETH balance accounting invariant

venimir-ssv:


yurii-ssv (approved):

mtabasco (changes requested):
@venimir-ssv 
Findings

Medium: the FUZZ-5 invariant is implemented, but the harness does not exercise all of the ETH-moving paths the readiness item called out. MAINNET-READINESS.md (line 3025) explicitly names claimEthRewards, withdrawNetworkETHEarnings, and withdrawOperatorEarnings; in [SSVClustersEchidna.sol (line 648)]() the invariant includes ethDaoBalance and stakingEthPoolBalance, with helpers at [SSVClustersEchidna.sol (line 813)](), [SSVClustersEchidna.sol (line 825)](), and [SSVClustersEchidna.sol (line 829)](), but this harness has no action paths that actually drive those outflow functions. In practice that means the new property is strong for cluster liabilities, but it does not fuzz the “ETH drain via another function” scenarios that FUZZ-5 was meant to cover.

PR #512, opened 2026-03-08
Ab/update flows md

andrew-blox:


andrew-blox:
@mtabasco there are 4 to dos in there most are questions or suggestions, please update them if relevant or delete them if not needed

Also some things should be updated for ex. i put operators max at 3000 something like that but its not 3000 its some var name operatorMaxValidators idk something like that, few spots could use this sort of update i think

PR #513, opened 2026-03-09
fix(staking): use SafeERC20 in rescueERC20

venimir-ssv:


yurii-ssv:
I think that if we will start to use SafeERC20 its batter to use it for all token transfers, not only in the SSVStaking module. We have transfers in cluster and operator related logic too

yurii-ssv (approved):

mtabasco (approved):

PR #537, opened 2026-03-16
fix(staking): carry sync fee remainder across accEthPerShare updates

venimir-ssv:


mtabasco:
Discarded, reference to https://github.com/ssvlabs/ssv-network/blob/ssv-staking/ssv-review/planning/MAINNET-READINESS.md BUG-18

PR #574, opened 2026-04-01
Fuzz/cat 1 7 mixed fee operators

venimir-ssv:


test/ssv-fuzz-engine/migration-mixed-fee-operators-legacy.fuzz.ts
```diff
@@ -0,0 +1,188 @@
...
+              const vCount = BigInt(ctx.state.cluster.cluster.validatorCount);
+              for (let i = 0; i < ctx.state.operators.length; i++) {
+                const op = ctx.state.operators[i];
+                if (op.fee === 0n) {
+                  await expect(
+                    ctx.network.connect(op.owner).withdrawAllOperatorEarnings(op.id),
+                  ).to.be.revertedWithCustomError(ctx.network, Errors.INSUFFICIENT_BALANCE);
+                } else {
+                  const earningsBefore = BigInt(await ctx.views.getOperatorEarnings(op.id));
+                  expect(earningsBefore).to.be.greaterThan(
```
yurii-ssv: I think we should use strict assertion here. Since we know all formulas better to assert agains pre-celculated expected value

yurii-ssv (changes requested):

test/ssv-fuzz-engine/migration-zero-fee-operators-legacy.fuzz.ts
yurii-ssv, line 80: I think better to randomize this value to cover move scenarios

yurii-ssv (approved):

PR #601, opened 2026-04-06
initial stress test

andrew-blox:


test/stress/SPEC.md
```diff
@@ -0,0 +1,1609 @@
...
+   - TS: clear pending fee fields
+   - After this, all clusters using this operator have changed burn rates
+
+4. TS must recompute all cluster burn rates using the new fee from this block forward.
+
+---
+
+## Staking Pool Sync Points
+
+The following contract functions call `_syncFees` internally, transferring accrued DAO ETH into the staking pool. TS must mirror this on every call:
```
mtabasco, line 1146: It looks like the `_syncFees` in the stress spec is modeling the staking/DAO accounting incorrectly.

The current text in `test/stress/SPEC.md` (line 1146) treats _syncFees as if it:

- moves newly accrued ETH out of ethDaoBalance
- zeroes ethDaoBalance
- separately accumulates a staking-pool bucket that must be added on top of network earnings in the ETH conservation invariant

That is not the protocol math. `_syncFees` snapshots `sp.ethDaoBalance = sp.networkTotalEarnings()` and sets `s.stakingEthPoolBalance = current`; it does not zero the DAO-side accounting bucket. The current stress invariant at `test/stress/SPEC.md` (line 1016) therefore double-counts the staking portion by summing:

- networkEarnings
- stakingPool
- stakerPending

The stress test’s critical rule is that getters are never the source of truth, so the TS model needs to mirror the contract math exactly and only use getters for verification afterward. For this area, TS should:

- compute current = networkTotalEarnings(...) from tracked protocol state
- compute newFees = current - previousStakingEthPoolBalance
- if totalStaked > 0, increase accEthPerShare by newFees / totalStaked
- set stakingEthPoolBalance = current
- keep ethDaoBalance as the checkpointed packed value used by networkTotalEarnings, not zero it

I’d update both the `_syncFees` pseudocode and the ETH conservation section so the TS reference math matches the real protocol instead of treating staking pool and DAO earnings as separate additive claim buckets.`

test/stress/SPEC.md
```diff
@@ -0,0 +1,1609 @@
...
+| `liquidate` | 6 | Only if `isLiquidatable` — see Liquidation section |
+| `updateClusterBalance` (oracle EB update) | 5 | See EB section — can cause liquidations (acceptable) |
+| `migrateClusterToETH` | 2 | Remaining SSV clusters |
+
+### SSV Legacy Cluster Operations (intentionally small pool — these should mostly fail)
+
+| Action | Weight | Notes |
+|---|---|---|
+| `registerValidator` on SSV cluster | 1 | **MUST revert** `IncorrectClusterVersion` |
+| `depositSSV` | 2 | Valid — SSV clusters can still receive deposits |
```
mtabasco, line 664: Are `depositSSV` and `withdrawSSV` only used on the pre-migration window?

The actual cluster deposit/withdraw paths validate `VERSION_ETH` only at `SSVClusters.sol` (line 186) and `SSVClusters.sol` (line 206)

test/stress/SPEC.md
```diff
@@ -0,0 +1,1609 @@
...
+  assert(await views.isLiquidated(owner, operatorIds, csCluster) === !cluster.active);
+
+  // 4. Validators (sample: check all registered + recently removed)
+  for (const [pk, registered] of cluster.validators) {
+    assert(await views.getValidator(owner, hexToBytes(pk)) === registered);
+  }
+
+  // 5. Effective balance (if explicit EB)
+  if (cluster.ebVUnits > 0n) {
+    const expectedEbPerValidator = cluster.ebVUnits * 32n / BPS_DENOMINATOR / BigInt(cluster.validatorCount);
```
mtabasco, line 887: `getEffectiveBalance` is being verified against the wrong TS quantity in `test/stress/SPEC.md` (line 885).

The current check computes:
```ts
const expectedEbPerValidator =
  cluster.ebVUnits * 32n / BPS_DENOMINATOR / BigInt(cluster.validatorCount);
```
That formula converts vUnits back to ETH and then divides by validatorCount, so it is computing an average per-validator EB. But the stress spec already defines effectiveBalance as the total cluster EB in ETH at `test/stress/SPEC.md` (line 804), and the contract getter returns that total value as well in `SSVViews.sol` (line 438).

Under the stress test’s critical rule, TS needs to compute the real expected contract value first and only then use the getter for verification. For this getter, the TS reference value should be total cluster EB:
```ts
const expectedEffectiveBalance =
  cluster.ebVUnits > 0n
    ? (cluster.ebVUnits * 32n) / BPS_DENOMINATOR
    : cluster.validatorCount * 32n;
```
Comparing against per-validator EB will fail for any multi-validator cluster, and it can also floor incorrectly when total EB is not evenly divisible by validator count.`

test/stress/SPEC.md
```diff
@@ -0,0 +1,1609 @@
...
+Called after **every write transaction** (and after every same-block batch). This is the heart of correctness verification.
+
+### Cluster Checks
+
+```typescript
+for (const [clusterId, cluster] of tsState.clusters) {
+  const { owner, operatorIds } = cluster;
+  const csCluster = toContractCluster(cluster);
+
+  // 1. Balance
```
mtabasco, line 868: checkState is calling the wrong getters for some cluster states.

In `test/stress/SPEC.md` (line 863), the pseudocode starts by doing:
```ts
const expectedBalance = computeClusterBalance(cluster, currentBlock, proto, ops);
const onChain = await views.getBalance(owner, operatorIds, csCluster);
assert(onChain === expectedBalance);
```
before it branches on cluster version later in the block.

That does not match the actual view API behavior:

- `getBalance(...)` is ETH-cluster only and also requires the cluster to be non-liquidated, see `SSVViews.sol `(line 389)
- `getBalanceSSV(...)` is the separate legacy SSV path, see `SSVViews.sol` (line 414)

So as written, the pseudocode would try to call getBalance(...) on:
- legacy SSV clusters
- liquidated ETH clusters
and those are not valid checks.

Under the stress test’s critical rule, TS should first compute the expected value from tracked state, then choose the matching getter only for states where that getter is actually defined. checkState should branch by cluster version and active/liquidated status before making the view call, roughly:
```ts 
if (cluster.version === VERSION_ETH) {
  if (cluster.active) {
    const expectedBalance = computeClusterBalance(cluster, currentBlock, proto, ops);
    assert(await views.getBalance(owner, operatorIds, csCluster) === expectedBalance);
  }
} else if (cluster.version === VERSION_SSV) {
  if (cluster.active) {
    const expectedBalance = computeClusterBalanceSSV(cluster, currentBlock, proto);
    assert(await views.getBalanceSSV(owner, operatorIds, csCluster) === expectedBalance);
  }
}
```
Same issue applies to other ETH/active-only getters in the same pseudocode block. The main fix is to make checkState select the getter based on the cluster state, not assume one getter is valid for every cluster record.

mtabasco: I checked that **the implementation is OK**, but the spec should be updated to reflect the actual branching in `checkState.ts`

test/stress/SPEC.md
mtabasco: Checked the code. The result is:

- The markdown spec for issue 1 is wrong.
- The actual stress engine code **already follows the correct compressed model**.
- The main remaining discrepancy is documentation: `SPEC.md` (line 1144) and the related invariant text should be updated to match `state.ts` (line 149) and `checkState.ts `(line 141).

One small nuance: the code is not a literal struct-for-struct mirror of the Solidity storage layout, but it is modeling the same economic state transition correctly for `_syncFees`.

test/stress/SPEC.md
mtabasco: Checked the code.

The real stress engine already uses the correct semantics:

- total cluster EB in ETH
- not average EB per validator

The bad formula only appears in the spec pseudocode at `test/stress/SPEC.md` (line 887), where it divides by validatorCount and turns total EB into per-validator EB.

mtabasco:
Closed as it is included in the hard-tests repo.

PR #602, opened 2026-04-07
[just] attach-module recipe uses env param

mtabasco:
`env` param added to `attach-module` recipe to require the correct target `SSVNetwork` proxy and update the latest deployment file.


venimir-ssv (approved):
I would add explicit confirmation flow y/n with tx details just before broadcast

yurii-ssv (approved):

PR #604, opened 2026-04-09
Test/improve echidna

mtabasco:
| ID | Description | Status |
|---|---|---|
| ECHIDNA-1 | Removed the disabled `echidna_liquidation_clears_eb_snapshot` property from `SSVClustersEchidna.sol` and deleted its stale README references. | ✅ Fixed |
| ECHIDNA-2 | Added coverage for the `MustUseLatestRoot` rule so `updateClusterBalance` must use `blockNum == latestCommittedBlock`, not merely an existing committed root. | ✅ Fixed |
| ECHIDNA-3 | Reworked the cluster coverage assessment to map `SSVClusters.sol` function-by-function to the harnesses that exercise it, instead of relying on raw file-level coverage alone. | ✅ Fixed |
| ECHIDNA-4 | Added migration-harness coverage for `migrateClusterToETH` when the legacy SSV cluster is already liquidated (`isLiquidated == true`). | ✅ Fixed |
| ECHIDNA-5 | Added `SSVClustersEchidna` coverage for successful `updateClusterBalance` on inactive/liquidated ETH clusters, asserting the call updates only the EB snapshot and does not mutate accounting or cluster state. | ✅ Fixed |
| ECHIDNA-6 | Added `SSVMigrationEchidna` coverage for the legacy `VERSION_SSV` branch of `updateClusterBalance`, asserting the call updates only the EB snapshot and does not mutate SSV cluster, DAO, or operator accounting state. | ✅ Fixed |
| ECHIDNA-7 | Added `bulkRegisterValidator` action support to `SSVValidatorsEchidna`, so existing validator/cluster/operator-count invariants now exercise both single and bulk registration paths. | ✅ Fixed |
| ECHIDNA-8 | Added `bulkRemoveValidator` action support to `SSVValidatorsEchidna`, so the existing validator, cluster, and operator-count invariants now also exercise bulk removal paths. | ✅ Fixed |
| ECHIDNA-9 | Added `bulkExitValidator` and unauthorized bulk-exit action support to `SSVValidatorsEchidna`, so the existing owner-only-exit and validator-consistency invariants now also exercise bulk exit paths. | ✅ Fixed |
| ECHIDNA-10 | Added `SSVWhitelistValidatorsEchidna` to cover batch-1 private-operator registration scenarios: mixed public/private clusters with a zero-fee private operator, whitelist-contract registration, and legacy SSV private operators initialized into ETH logic while preserving whitelist rules. | ✅ Fixed |
| ECHIDNA-11 | Extended `SSVWhitelistValidatorsEchidna` so the same private-operator authorization and accounting checks now also exercise `bulkRegisterValidator` for direct whitelist, whitelist-contract, and legacy-private ETH-init flows. | ✅ Fixed |
| ECHIDNA-12 | Added `SSVAccountingEchidna` coverage for `withdrawNetworkSSVEarnings` in nonzero-accrual states, asserting exact `daoBalance` decrement, `daoIndexBlockNumber` checkpoint reset, and over-withdraw reverts; the real token-outflow path remains covered in `SSVDAOEchidna` via an external caller. | ✅ Fixed |
| ECHIDNA-13 | Added a targeted `migrateClusterToETH` regression test showing the production SSV refund path does not invoke owner callbacks: a contract owner with a reverting `receive()` still migrates successfully and receives the SSV refund. This closes the receiver-side reentrancy concern for migration without modeling non-production tokens. | ✅ Fixed |
| ECHIDNA-14 | Added `SSVLegacyValidatorRemovalEchidna` to cover the legacy `VERSION_SSV` remove surface from the spec and flows: seeded active and already-liquidated `removeValidator` / `bulkRemoveValidator` states, exact SSV operator/DAO count settlement on active removal, and the rule that liquidated-cluster removal must not decrement counts twice after liquidation. | ✅ Fixed |
| ECHIDNA-15 | Extended `SSVOperatorsEchidna` with direct coverage for the legacy `reduceOperatorFee -> ensureETHDefaults()` branch and for `withdrawAllVersionOperatorEarnings`, reusing the existing reduce/withdraw invariants to check ETH-default initialization, DAO minimum-fee respect, dual-balance zeroing, and combined ETH+SSV payout accounting. | ✅ Fixed |
| ECHIDNA-16 | Extended `SSVWhitelistValidatorsEchidna` with private-operator follow-on flows after first use: whitelist mutation plus fee change on an in-use private operator, and public/private privacy toggles on an in-use private operator, asserting authorization updates take effect without breaking validator-count or cluster-hash accounting. | ✅ Fixed |


test/echidna/SSVValidatorsEchidna.sol
```diff
@@ -165,6 +225,83 @@ contract SSVValidatorsEchidna is SSVValidators {
         } catch {}
     }
 
+    function action_bulk_remove(uint256 seed) external {
```
venimir-ssv, line 261: I may be missing another harness but I don’t see an unauthorized bulk-remove path being exercised here echidna_owner_only_remove() still appears to be driven only by the single-remove unauthorized action, so I think bulkRemoveValidator() needs its own negative-path coverage

test/echidna/SSVValidatorsEchidna.sol
```diff
@@ -139,6 +160,45 @@ contract SSVValidatorsEchidna is SSVValidators {
         } catch {}
     }
 
+    function action_bulk_register(uint256 seed, uint8 ownerSeed, uint8 operatorsSeed) external {
```
venimir-ssv, line 163: This expands bulk registration coverage but I don’t see a case that deliberately sends duplicate keys through bulkRegisterValidator() the single-register path probes duplicates explicitly this bulk path seems to return before making the call when it encounters an already-known key I think it would be worth adding a direct duplicate batch regression

test/echidna/SSVValidatorsEchidna.sol
mtabasco: You are correct, Im adding `action_bulk_remove_unauthorized` , thanks!

test/echidna/SSVValidatorsEchidna.sol
mtabasco: fixed!

venimir-ssv (approved):

PR #605, opened 2026-04-10
feat: cat 4-5-6 fuzz tests

yurii-ssv:


test/ssv-fuzz-engine/core/setup.ts
```diff
@@ -34,14 +33,15 @@ export async function registerFuzzOperators(
   owner: HardhatEthersSigner,
   count: number,
   fees: bigint[],
-  setPrivate: boolean = true,
+  keyOffset: number = 1000,
```
venimir-ssv, line 36: The old 5th argument was the boolean privacy flag, but now it became keyOffset, while the underlying contract API still uses registerOperator(publicKey, fee, setPrivate) in ISSVOperators.sol (line 122). We still have existing suites calling registerFuzzOperators(..., false) with the old meaning, for example eth-withdraw-boundary.fuzz.ts (line 44).

I think we need either a backwards-compatible helper signature, or all related call sites should be updated in the same PR

test/ssv-fuzz-engine/eb-multi-cluster-shared-ops.fuzz.ts
```diff
@@ -1,400 +1,237 @@
...
 
-function computeExpectedBurnRate(
-  clusterOps: OperatorRecord[],
-  networkFee: bigint,
-  vUnits: bigint,
-): bigint {
-  let opFeeSum = 0n;
-  for (const op of clusterOps) {
-    opFeeSum += op.fee;
+async function lifecycle(ctx: FuzzContext<State>): Promise<void> {
```
venimir-ssv, line 51: I may be missing another suite, but this rewrite seems to drop the explicit operator/network earnings isolation checks that the previous version had. The new lifecycle still checks burn-rate isolation and validator counts, but I no longer see the getOperatorEarnings() / getNetworkEarnings() delta proofs here. Was that coverage moved elsewhere, or should part of it stay in this test?

mtabasco (changes requested):
# PR #605 Evaluation: CAT-4, CAT-5, CAT-6 Fuzz Tests

## Overview

PR #605 adds **16 new fuzz test files** and makes significant infrastructure improvements (assertions, fuzz-helpers, setup refactored to use vUnits natively). The tests target operator lifecycle (CAT-4), concurrent mixed operations (CAT-5), and governance under load (CAT-6).

---

## Coverage Matrix: Plan vs. Implementation

### CAT-4: Operator Lifecycle (8 planned, 6 implemented)

| Scenario | File | Status | Notes |
|---|---|---|---|
| **CAT-4-1** (ensureETHDefaults) | — | **MISSING** | No dedicated test. Partially covered by `migration-shared-operators.fuzz.ts` which verifies `OperatorFeeExecuted` events on migration, but doesn't test the zero-fee→zero-ETH-fee path separately |
| **CAT-4-2** (declare→execute→cluster) | `operator-fee-declare-execute-lifecycle.fuzz.ts` | **GOOD** | Full lifecycle: accrue→declare→verify old fee→execute too early revert→execute in window→verify new burn rate |
| **CAT-4-3** (fee reduce immediate) | — | **MISSING** | No dedicated test. `chaos-actions.ts` includes `reduceOperatorFee` but doesn't verify immediate application or reduce-to-zero path |
| **CAT-4-4** (removal: earnings + cluster) | — | **PARTIALLY COVERED** | `operator-removal-interleave.fuzz.ts` verifies removal but doesn't verify final SSV/ETH snapshot settlement or "all earnings auto-withdrawn to owner" per plan |
| **CAT-4-5** (removal: multi-cluster) | `operator-removal-multi-cluster.fuzz.ts` | **GOOD** | Tests shared operator removal across 2 clusters, verifies both clusters affected, validator registration reverts |
| **CAT-4-6** (dual ETH+SSV withdrawal) | `operator-dual-earnings-withdrawal.fuzz.ts` | **GOOD** | Full lifecycle with legacy→migrated operators. Tests withdrawETH, withdrawSSV, withdrawAllVersion, ETH-only SSV revert |
| **CAT-4-7** (declare overwrite) | `operator-fee-declare-overwrite.fuzz.ts` | **GOOD** | Declares fee_A, then fee_B, verifies only fee_B is executed |
| **CAT-4-8** (fee increase limit) | `operator-fee-increase-limit.fuzz.ts` | **GOOD** | Declares at max → cancels → declares above max → verifies revert |

### CAT-5: Concurrent Mixed Operations (6 planned, 5 implemented)

| Scenario | File | Status | Notes |
|---|---|---|---|
| **CAT-5-1** (multi-owner interleaved) | `chaos.fuzz.ts` + `chaos-actions.ts` | **GOOD** | Comprehensive chaos test with 17 weighted random actions across 2 clusters with shared operators. Verifies ETH conservation after every action |
| **CAT-5-2** (migration race) | `migration-shared-operators.fuzz.ts` | **GOOD** | Two SSV clusters migrate sequentially. Verifies ensureETHDefaults idempotency: first migration emits N events, second only emits for unique operators |
| **CAT-5-3** (removal during cluster ops) | `operator-removal-interleave.fuzz.ts` | **GOOD** | Randomizes execution order of {withdraw, removeOperator, registerValidator}. Verifies post-removal registerValidator reverts |
| **CAT-5-4** (EB update→liquidation→reactivation) | `eb-auto-liquidation-lifecycle.fuzz.ts` | **GOOD** | Full lifecycle: high EB triggers auto-liquidation → reactivate with stale EB → new EB update → survival |
| **CAT-5-5** (concurrent SSV + ETH) | `dual-version-coexistence.fuzz.ts` | **GOOD** | SSV cluster + ETH cluster on shared operators. Verifies dual tracking (INV-10), version exclusivity (INV-9), migration accounting |
| **CAT-5-6** (multiple liquidations same block) | `multi-liquidation-same-block.fuzz.ts` | **GOOD** | Uses `evm_setAutomine(false)` to mine both liquidations in same block. Verifies bounty conservation |

### CAT-6: Governance Under Load (7 planned, 4 implemented)

| Scenario | File | Status | Notes |
|---|---|---|---|
| **CAT-6-1** (network fee change) | `network-fee-dual-cluster.fuzz.ts` | **PARTIAL** | Tests fee change + cluster balance impact for both ETH and SSV clusters. But doesn't explicitly verify "decreased faster / slower" as plan requires — verifies exact balance computation instead (arguably stronger) |
| **CAT-6-2** (liquidation threshold change) | `chaos-actions.ts` only | **WEAK** | Only exercised randomly in chaos test via `changeLiquidationThreshold`. No dedicated test verifying that an existing cluster transitions from non-liquidatable to liquidatable after threshold increase |
| **CAT-6-3** (min collateral change) | `chaos-actions.ts` only | **WEAK** | Same as above — random chaos action only, no targeted verification |
| **CAT-6-4** (max fee blocks pending declaration) | `governance-max-fee-pending-declaration.fuzz.ts` | **GOOD** | Declares fee → advances to window → lowers max → execute reverts |
| **CAT-6-5** (oracle replacement during voting) | `oracle-replacement.fuzz.ts` | **GOOD** | Comprehensive chaos: replace during partial quorum, verify old oracle rejected, double-vote rejected, non-oracle rejected, new oracle can complete quorum |
| **CAT-6-6** (min operator ETH fee change) | — | **MISSING** | No test. Plan requires verifying existing operators not retroactively changed, new declarations enforced against new minimum, reduce below new min reverts (except to 0) |
| **CAT-6-7** (dual SSV+ETH network fee change) | `network-fee-dual-cluster.fuzz.ts` | **GOOD** | Randomizes both ETH and SSV fee changes, verifies fee independence and SSV conservation |

---

## Invariant Coverage from SPEC §11

| Invariant | Coverage | Where |
|---|---|---|
| **INV-1** (ETH Conservation) | **STRONG** | `checkChaosInvariants` in chaos-actions.ts, `assertEthConservation` in several tests |
| **INV-2** (SSV Conservation) | **GOOD** | New `assertSSVConservation` assertion, used in dual-version and network-fee tests |
| **INV-3** (Validator Count) | **GOOD** | `checkChaosInvariants` verifies per-operator and network-wide validator counts |
| **INV-4** (vUnit Consistency) | **PARTIAL** | Only verified indirectly via burn rate matching. No explicit `daoTotalEthVUnits` check in new tests (was in older tests via `assertDaoVUnitsMatchCluster`) |
| **INV-5** (Cluster Hash) | Implicit | Validated by every cluster operation succeeding (stale hash = revert) |
| **INV-6** (cSSV Supply) | Not tested | Out of scope for CAT-4/5/6 |
| **INV-7** (Accumulator Monotonicity) | Not tested | Out of scope for CAT-4/5/6 |
| **INV-8** (Oracle Monotonicity) | **GOOD** | Block number guards in all oracle tests |
| **INV-9** (Version Exclusivity) | **GOOD** | New `assertVersionExclusivity` assertion used in dual-version test |
| **INV-10** (Operator Dual Tracking) | **GOOD** | `checkDualTracking` in dual-version-coexistence.fuzz.ts |

---

## Findings

### F-1 [Medium]: `assertOperatorEarnings` hardcodes implicit vUnits — misses explicit EB cases

In `assertions.ts:114`:
```ts
ctx.state.lastOperatorEarnings = { block, earnings: currentEarnings, vUnits: BigInt(cluster.cluster.validatorCount) * BPS_DENOMINATOR };
```

This always computes vUnits as `validatorCount * 10,000` (implicit EB). But per SPEC §10 "ETH Operator Earnings (with EB)":
```
effectiveVUnits = seb.operatorEthVUnits[operatorId] + operator.ethValidatorCount * BPS_DENOMINATOR
```

For clusters with **explicit EB** (where `clusterEB.vUnits != 0`), the actual vUnits include the deviation from the oracle. The assertion will produce **wrong expected deltas** for any test that uses explicit EB and then calls `assertOperatorEarnings`. Currently the new tests don't use this function with explicit EB, but it's a correctness bug in shared infrastructure that was widened by this PR's refactor (the old code had the same issue with `validatorCount`).

Similarly, `assertClusterBalance`, `assertNetworkEarnings`, and all `Phase*` variants have this same pattern. The older `assertOperatorEarningsWithEB` and `assertClusterBalanceWithEB` correctly accept vUnits from state — these should be used for EB-enabled clusters.

### F-2 [Low]: `operator-dual-earnings-withdrawal` has off-by-one-block accounting assumption

In the `withdraw-all-version` phase:
```ts
const migrationValidatorCount = 2n;
const oneBlockEthDelta = DEFAULT_OPERATOR_ETH_FEE * migrationValidatorCount;
expect(ethWithdrawn).to.equal(expectedEthEarnings + oneBlockEthDelta);
```

This hardcodes a **1-block ETH earning delta** between querying `getOperatorEarnings` and the `withdrawAllVersionOperatorEarnings` tx executing. This is fragile — if the test timing changes or if the `ticks` configuration causes the query and withdrawal to land in the same block, this will fail. The `migrationValidatorCount = 2n` is also hardcoded but the test plan says CAT-4-6 uses O8 operators with legacy SSV + migrated ETH clusters with fuzzed validator counts.

**Recommendation:** Query earnings **in the same block** as the withdrawal (via `staticCall`) or use `evm_setAutomine(false)` to ensure deterministic block distance.

### F-3 [Low]: `checkChaosInvariants` checks `getOperatorById().validatorCount` but this is ETH-only

In chaos-actions.ts line ~663:
```ts
expect(BigInt((await ctx.views.getOperatorById(op.id)).validatorCount)).to.equal(expected);
```

`getOperatorById` returns `ethValidatorCount`. Since the chaos test only operates ETH clusters, this is correct. But it's worth noting for anyone extending chaos to include SSV clusters — `getOperatorByIdSSV` would be needed for dual tracking.

### F-4 [Info]: Regression in assertions — removed `assertDaoVUnitsMatchCluster`

The refactored `eb-boundary-validation` and `eb-implicit-to-explicit` tests **removed** calls to `assertDaoVUnitsMatchCluster` (which verified INV-4 explicitly). The new versions verify burn rate matches expected vUnits (functionally equivalent for correctness), but INV-4 (`daoTotalEthVUnits` matching) is no longer explicitly verified in these test files. The invariant is still checked elsewhere in older tests that weren't changed.

### F-5 [Info]: Missing tests for 3 plan items

1. **CAT-4-1** (ensureETHDefaults for O1/O2 operators): Needs a dedicated test showing zero-SSV-fee → zero-ETH-fee mapping and non-zero-SSV-fee → DEFAULT_OPERATOR_ETH_FEE mapping with event verification
2. **CAT-4-3** (reduceOperatorFee immediate): Needs a dedicated test verifying immediate application (no timelock), snapshot settlement at old fee, and reduce-to-zero 
3. **CAT-6-6** (min operator ETH fee change): Needs test verifying existing operators not retroactively changed, new declarations enforced, reduce below new min reverts (except 0)

### F-6 [Info]: `computeSSVClusterBalance` added but only used in `network-fee-dual-cluster.fuzz.ts`

The new `computeSSVClusterBalance` helper in fuzz-helpers.ts correctly uses `DEDUCTED_DIGITS` instead of `ETH_DEDUCTED_DIGITS` for legacy SSV accounting. Good addition — the old `computeClusterBalance` function was ETH-only. This should be used consistently in any future SSV cluster balance checks.

---

## Infrastructure Quality Assessment

**Positives:**
- The `chaos-actions.ts` engine is well-designed: weighted random actions, `tryAction` for soft failures, and comprehensive invariant checking after every action
- Consistent use of `FuzzContext<State>` pattern with typed phase machines
- `computeMaxAllowedFee` correctly implements ceiling division matching the contract's `OperatorLib.checkFeeIncrease`
- `parseFeeExecutedEvents` replaces the previous `collectOperatorFeeExecutedEvents` with a cleaner implementation
- `assertSSVConservation` and `assertVersionExclusivity` are good additions for invariant checking

**Concerns:**
- The refactor from `validatorCount` to `vUnits` in all assertion functions is correct for the narrow case (implicit EB only), but creates a false sense of generality — the assertions still hardcode `vUnits = validatorCount * BPS_DENOMINATOR` and don't accept external vUnits
- `registerFuzzOperators` parameter change from `setPrivate: boolean` to `keyOffset: number, isETHOperator: boolean` is a breaking change but properly migrated across all callers

---

## Summary

| Metric | Status |
|---|---|
| Plan coverage | **17/21** scenarios implemented (81%) |
| Spec invariant coverage | 7/10 relevant invariants checked |
| New test files | 16 added |
| Lines added | ~3,200 |
| Infrastructure quality | Good, with noted vUnit generality concern |
| Blocking issues | F-1 (assertion correctness for explicit EB), F-2 (fragile accounting) |
| Missing items | CAT-4-1, CAT-4-3, CAT-6-2 (dedicated), CAT-6-3 (dedicated), CAT-6-6 |

**Verdict:** Solid PR with good coverage of the most complex scenarios (chaos, dual-version, oracle replacement, same-block liquidations). The missing scenarios are all relatively straightforward to add. F-1 should be addressed to prevent future misuse of assertions with explicit-EB clusters.

mtabasco:
Amazing work @yurii-ssv !

Guys @venimir-ssv @yurii-ssv Let's complete the pending scenarios together. When you pick an item, please share in our channel. thanks!!

I half-trusted the recommendations; I think the most meaningful one for now is the first one.

venimir-ssv (approved):

mtabasco (approved):

